PR #10, opened 2021-10-18
[PAYG-38] Add csharp TrueLayer.Signing lib

alexheretic:
# TrueLayer.Signing
C# library to produce & verify TrueLayer API requests signatures.

```xml
<PackageReference Include="TrueLayer.Signing" Version="_" />
```

```csharp
using TrueLayer.Signing;

var tlSignature = Signer.SignWithPem(kid, privateKey)
    .Method("POST")
    .Path(path)
    .Header("Idempotency-Key", idempotency_key)
    .Body(body)
    .Sign();
```

## Implementation notes
* ~I used `net5.0` runtime instead of LTS 3.1 because of the [`ImportFromPem` support](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.rsa.importfrompem?view=net-5.0).~ Supports `netstandard2.1` & `net5.0`.
* Signing/verifying v1 body-only signatures is not supported. We could add this, similar to the rust lib, if desired. However, this shouldn't be necessary (from a signer pov) with paydirect-api v1 forwards supporting full request signatures.

### DIFF
--- a/csharp/src/Signer.cs
+++ b/csharp/src/Signer.cs
@@ -17,14 +17,19 @@ namespace TrueLayer.Signing
         /// RFC 7468 PEM-encoded data and the key's kid.
         /// </summary>
         public static Signer SignWithPem(string kid, ReadOnlySpan<char> privateKeyPem)
-            => new Signer(kid, privateKeyPem);
+            => SignWith(kid, privateKeyPem.ParsePem());
 
         /// <summary>
         /// Start building a request Tl-Signature header value using private key
         /// RFC 7468 PEM-encoded data and the key's kid.
         /// </summary>
         public static Signer SignWithPem(string kid, ReadOnlySpan<byte> privateKeyPem)
-            => new Signer(kid, Encoding.UTF8.GetString(privateKeyPem));
+            => SignWithPem(kid, Encoding.UTF8.GetString(privateKeyPem));
+
+        /// <summary>
+        /// Start building a request Tl-Signature header value using private key and the key's kid.
+        /// </summary>
+        public static Signer SignWith(string kid, ECDsa privateKey) => new Signer(kid, privateKey);
 
         private ECDsa key;
         private string kid;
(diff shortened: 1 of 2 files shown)


csharp/src/truelayer-signing.csproj
```diff
@@ -0,0 +1,17 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <Version>0.1.0</Version>
+    <TargetFramework>net5.0</TargetFramework>
+    <DocumentationFile>bin\$(Configuration)\$(TargetFramework)\$(AssemblyName).xml</DocumentationFile>
+    <Nullable>enable</Nullable>
+    <WarningsAsErrors>CS8600;CS8602;CS8603;CS8604;CS8618</WarningsAsErrors>
+    <Authors>Alex Butler</Authors>
```
tl-mauro-franchi, line 8: I think we should put `TrueLayer` here as we publish it from our Org 🤔

tl-mauro-franchi: Also maybe add:
```
<Company>TrueLayer</Company>
<Copyright>TrueLayer (c) $([System.DateTime]::Now.ToString(yyyy))</Copyright>
```
And the github info tags:
```
<PackageProjectUrl>https://github.com/truelayer/{repository}</PackageProjectUrl>
    <PackageReleaseNotes>See $(PackageProjectUrl)/releases for details.</PackageReleaseNotes>
    <PackageRequireLicenseAcceptance>false</PackageRequireLicenseAcceptance>
    <Repository>truelayer-signing</Repository>
    <RepositoryType>git</RepositoryType>
    <RepositoryUrl>$(PackageProjectUrl).git</RepositoryUrl>
```

alexheretic: Thanks I left my name in authors, as from what I can see this is generally done. But I don't mind too much either way.

PR #14, opened 2021-11-02
[WIP] Move JS library to TS

tl-matt-davis:
This PR migrates the javascript library from JS to TS.

The idea here is that we can emit the JS lib with TS types built in, so that JS consumers are unchanged, but TS consumers don't need a separate types declaration.

We're trying to do this with no breaking changes, and still including the original JSDoc typings.


alexheretic:
Wait is this really needed, doesn't the jsdoc do enough? It's not as if there are any complex types involved.

tl-matt-davis:
@alexheretic you can't import the JS library into a TS project, unless you provide type definitions manually or enable importing of JS in your tsconfig. I'd say for external facing libraries this is a bit of friction that we should for better developer experience.

So normally providing for TS is done in two ways:

1. Publish two packages, one for JS, one as a standalone types package (in this case @types/truelayer-signing) which you add to https://github.com/DefinitelyTyped/DefinitelyTyped

2. You write your library with TS, and generate JS build output + TS type definitions, releasing both together. JS doesn't read these files, so there's no difference to JS consumers

Option #2 is generally industry standard - and makes maintenance easier as we don't have to worry about maintaining a separate package on DefinitelyTyped.

alexheretic:
Ok that makes sense. I think we should include a plain javascript usage test too. Perhaps just the original `sign should sign a full request...` test will be enough there.

alexheretic:
We could probably rename "javascript" -> "nodejs" in the readmes too. And mention the lib supports ts & js.

alexheretic:
Fyi https://github.com/TrueLayer/truelayer-signing/pull/16 includes new scenario tests for nodejs & support for no-header signatures in the verifier. Should be easy enough to merge I think!

alexheretic:
Sorry another conflict because I updated the changelog. We should add a point about ts support in the changelog too

tl-peter-mills:
@alexheretic Should I increment version to `0.2.0` and update the changelog?

alexheretic:
> @alexheretic Should I increment version to `0.2.0` and update the changelog?

I don't think this _is_ a breaking change right, the js usage will continue to work without any change right? As such just a non-breaking `0.1.1` release will cover it, even if it's adding a major feature.

You can just use `## Unreleased` for now too, though same difference as I'll push a release out once it's merged :+1:

nodejs/src/lib.ts
```diff
@@ -213,13 +213,8 @@ function verify(args: VerifyParameters) {
...
   return parseSignature(tlSignature).headerJson.kid;
 }
 
-export default {
-  sign,
-  verify,
-  extractKid,
-  SignatureError,
-};
+export { SignatureError };
```
alexheretic: When I added a plain js usage test I saw that the ts conversion had actually broken import usage from js.

I fixed it so that imports would be unchanged. However, this means in typescript imports will look like
```typescript
import * as tlSigning from 'truelayer-signing';
```

I'm not sure if that's an issue really, but if we want to change it it'll be a breaking change.

nodejs/tsconfig.json
```diff
@@ -6,8 +6,9 @@
     "outDir": "./lib",
     "strict": true,
     "resolveJsonModule": true,
+    "allowJs": true,
```
alexheretic, line 9: I added this to allow the .js usage test to run without warnings. I'm not sure if there's a less blanket way?

alexheretic (approved):
If you guys are cool with the import changes I made I think we're good to go :+1:

nodejs/tsconfig.json
tl-peter-mills: Thanks for adding the test! I've added the js test to the `exclude` array in the `tsconfig`, which means we can remove the `allowJs` setting

tl-peter-mills:
I've got no problem with the import changes. I've made a few small changes since your review, let me know if you're happy and then we're good to merge!

alexheretic (approved):
lgtm great work!

alexheretic:
Published on npm as **v0.1.1**

PR #15, opened 2021-11-04
TrueLayer Signing PHP

stefandanaita:


php/tests/VerifierTest.php
```diff
@@ -0,0 +1,74 @@
...
+        ->path('/test')
+        ->header('X-Idempotency-Key', 'idempotency-test')
+        ->body('{"random-key": "random-value"}')
+        ->requireHeaders([
+            'X-Idempotency-Key'
+        ]);
+
+    expect($verifier->verify($signature))->not->toThrow(Exception::class);
+});
+
```
alexheretic, line 29: One important scenario is the static signature verify, see https://github.com/TrueLayer/truelayer-signing/blob/0d252d4906f77bce2aaa9352c30df20d729ad52f/rust/tests/usage.rs#L87..L102

If every lang ensure that this exact signature works for the verify impl then we ensure that all langs have consistent signing implementations. Otherwise it can be quite easy to make a signing impl that works, but isn't exactly the same as the other langs, and therefore won't actually work when it hits our rust/c# servers.

php/tests/VerifierTest.php
```diff
@@ -0,0 +1,74 @@
...
+it('should throw when required header is missing', function () {
```
alexheretic, line 30: Could do the other verify scenario like "method mismatch" etc, but not a huge deal.

php/composer.json
```diff
@@ -0,0 +1,33 @@
...
+    ],
+    "require": {
+        "php": "^7.4|^8.0",
+        "web-token/jwt-core": "^2.2",
+        "web-token/jwt-signature": "^2.2",
+        "web-token/jwt-signature-algorithm-ecdsa": "^2.2",
+        "web-token/jwt-key-mgmt": "^2.2"
+    },
+    "require-dev": {
+        "pestphp/pest": "^1.20",
```
lboynton, line 30: I would recommend adding `phpstan` or `psalm` here, it's a static analysis tool for finding bugs in the code. See https://phpstan.org/user-guide/getting-started

I've just run phpstan at the strictest settings and it's found a few issues:

```
vendor/bin/phpstan analyse --level 9 src tests
 15/15 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%

 ------ ----------------------------------------------------------------------------------------------------------------------------- 
  Line   src/AbstractJws.php                                                                                                          
 ------ ----------------------------------------------------------------------------------------------------------------------------- 
  13     Property TrueLayer\Signing\AbstractJws::$request_headers type has no value type specified in iterable type array.            
         💡 See: https://phpstan.org/blog/solving-phpstan-no-value-type-specified-in-iterable-type                                     
  39     Method TrueLayer\Signing\AbstractJws::headers() has parameter $headers with no value type specified in iterable type array.  
         💡 See: https://phpstan.org/blog/solving-phpstan-no-value-type-specified-in-iterable-type                                     
 ------ ----------------------------------------------------------------------------------------------------------------------------- 

 ------ ------------------------------------------------------------------------------------------------------------------------------- 
  Line   src/Contracts/Jws.php                                                                                                          
 ------ ------------------------------------------------------------------------------------------------------------------------------- 
  11     Method TrueLayer\Signing\Contracts\Jws::headers() has parameter $headers with no value type specified in iterable type array.  
         💡 See: https://phpstan.org/blog/solving-phpstan-no-value-type-specified-in-iterable-type                                       
 ------ ------------------------------------------------------------------------------------------------------------------------------- 

 ------ ------------------------------------------------------------------------------------------------------------------------------------------- 
  Line   src/Contracts/Verifier.php                                                                                                                 
 ------ ------------------------------------------------------------------------------------------------------------------------------------------- 
  14     Method TrueLayer\Signing\Contracts\Verifier::requireHeaders() has parameter $headers with no value type specified in iterable type array.  
         💡 See: https://phpstan.org/blog/solving-phpstan-no-value-type-specified-in-iterable-type                                                   
 ------ ------------------------------------------------------------------------------------------------------------------------------------------- 

 ------ ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
  Line   src/Verifier.php                                                                                                                                                                                    
 ------ ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
  29     Property TrueLayer\Signing\Verifier::$required_headers type has no value type specified in iterable type array.                                                                                     
         💡 See: https://phpstan.org/blog/solving-phpstan-no-value-type-specified-in-iterable-type                                                                                                            
  66     Method TrueLayer\Signing\Verifier::requireHeaders() has parameter $headers with no value type specified in iterable type array.                                                                     
         💡 See: https://phpstan.org/blog/solving-phpstan-no-value-type-specified-in-iterable-type                                                                                                            
  81     PHPDoc tag @throws with type                                                                                                                                                                        
         TrueLayer\Signing\Exceptions\InvalidAlgorithmException|TrueLayer\Signing\Exceptions\InvalidSignatureException|TrueLayer\Signing\Exceptions\InvalidTrueLayerSignatureVersionException|TrueLayer\Sig  
         ning\Exceptions\RequestPathNotFoundException|TrueLayer\Signing\Exceptions\RequiredHeaderMissingException|TrueLayer\Signing\Exceptions\TrueLayerHeadersNotFoundException is not subtype of           
         Throwable                                                                                                                                                                                           
 ------ ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 

 ------ ---------------------------------------------------------------------------------------------------------------------------- 
  Line   tests/MockData.php                                                                                                          
 ------ ---------------------------------------------------------------------------------------------------------------------------- 
  8      Method TrueLayer\Signing\Tests\MockData::generateKeyPair() return type has no value type specified in iterable type array.  
         💡 See: https://phpstan.org/blog/solving-phpstan-no-value-type-specified-in-iterable-type                                    
 ------ ---------------------------------------------------------------------------------------------------------------------------- 

 ------ ---------------------------------------------------------------------------------------------- 
  Line   tests/SignerTest.php                                                                          
 ------ ---------------------------------------------------------------------------------------------- 
  9      Static method TrueLayer\Signing\Signer::signWithKey() invoked with 3 parameters, 2 required.  
  12     Access to an undefined property Pest\Expectation|Pest\Support\Extendable::$not.               
  13     Call to an undefined method Pest\Expectation|Pest\Support\Extendable::toBeString().           
  18     Static method TrueLayer\Signing\Signer::signWithKey() invoked with 3 parameters, 2 required.  
 ------ ---------------------------------------------------------------------------------------------- 

 ------ ---------------------------------------------------------------------------------------------- 
  Line   tests/VerifierTest.php                                                                        
 ------ ---------------------------------------------------------------------------------------------- 
  10     Static method TrueLayer\Signing\Signer::signWithKey() invoked with 3 parameters, 2 required.  
  13     Call to an undefined method TrueLayer\Signing\AbstractJws::sign().                            
  19     Call to an undefined method TrueLayer\Signing\AbstractJws::requireHeaders().                  
  27     Access to an undefined property Pest\Expectation|Pest\Support\Extendable::$not.               
  27     Result of method TrueLayer\Signing\Verifier::verify() (void) is used.                         
  32     Static method TrueLayer\Signing\Signer::signWithKey() invoked with 3 parameters, 2 required.  
  35     Call to an undefined method TrueLayer\Signing\AbstractJws::sign().                            
  41     Call to an undefined method TrueLayer\Signing\AbstractJws::requireHeaders().                  
  57     Static method TrueLayer\Signing\Signer::signWithKey() invoked with 3 parameters, 2 required.  
  60     Call to an undefined method TrueLayer\Signing\AbstractJws::sign().                            
  66     Call to an undefined method TrueLayer\Signing\AbstractJws::requireHeaders().                  
 ------ ---------------------------------------------------------------------------------------------- 

                                                                                                                        
 [ERROR] Found 23 errors                                                                                                
```

lboynton: Could also consider adding https://github.com/Roave/SecurityAdvisories as well

php/src/AbstractJws.php
```diff
@@ -0,0 +1,66 @@
+<?php
+
```
lboynton, line 2: The strict type declaration is needed at the top of each file.
```suggestion
declare(strict_types=1);

```

If you don't have it then types will be coerced. (At least on PHP 7.4, I'm pretty sure the same will happen on 8).

php/src/AbstractJws.php
```diff
@@ -0,0 +1,66 @@
...
+namespace TrueLayer\Signing;
+
+use TrueLayer\Signing\Contracts\Jws as IJws;
+use TrueLayer\Signing\Exceptions\RequestPathNotFoundException;
+
+abstract class AbstractJws implements IJws
+{
+    protected string $request_method = 'POST';
```
lboynton, line 10: Looks like there is a mixture of snake case and camel case in the code? Would recommend sticking to one, camel case would be my choice.

php/src/Verifier.php
```diff
@@ -0,0 +1,106 @@
...
+        if ($jwsHeaders['alg'] !== TrueLayerSignatures::ALGORITHM) {
+            throw new InvalidAlgorithmException();
+        }
+
+        if ($jwsHeaders['tl_version'] !== TrueLayerSignatures::SIGNING_VERSION) {
+            throw new InvalidTrueLayerSignatureVersionException();
+        }
+
+        foreach ($this->required_headers as $header) {
+            if (!in_array($header, $jwsHeaders['tl_headers'])) {
```
lboynton, line 97: Strictness
```suggestion
            if (!in_array($header, $jwsHeaders['tl_headers'], true)) {
```

php/src/Signer.php
```diff
@@ -0,0 +1,80 @@
...
+    public static function signWithPemFile(string $kid, string $path, ?string $passphrase): Signer
+    {
+        $jwk = JWKFactory::createFromKeyFile($path, $passphrase, [
+            'use' => 'sig',
+        ]);
+
+        return new self($kid, $jwk);
+    }
+
+    protected function __construct(string $kid, JWK $jwk)
```
lboynton, line 50:
```suggestion
    private function __construct(string $kid, JWK $jwk)
```
Makes a bit more sense for a final class?

alexheretic:
Note: Some more scenarios to test: https://github.com/TrueLayer/truelayer-signing/pull/16

stefandanaita:
> Note: Some more scenarios to test: #16

Cheers, will look and add what needs adding 🙏🏻

php/tests/VerifierTest.php
stefandanaita: Done!

php/tests/VerifierTest.php
stefandanaita: Done!

stefandanaita:
> Note: Some more scenarios to test: #16

Done both.

php/src/Contracts/Signer.php
```diff
@@ -0,0 +1,16 @@
...
+namespace TrueLayer\Signing\Contracts;
+
+use Jose\Component\Core\JWK;
+
+interface Signer extends Jws
+{
+    public static function signWithKey(string $kid, JWK $jwk): self;
+    public static function signWithPem(string $kid, string $pem, ?string $passphrase): self;
+    public static function signWithPemBase64(string $kid, string $pemBase64, ?string $passphrase): self;
+    public static function signWithPemFile(string $kid, string $path, ?string $passphrase): self;
```
lboynton, line 13: May be worth making the passphrase default to null so you don't have to set the last param to null yourself
```suggestion
    public static function signWithPem(string $kid, string $pem, ?string $passphrase = null): self;
    public static function signWithPemBase64(string $kid, string $pemBase64, ?string $passphrase = null): self;
    public static function signWithPemFile(string $kid, string $path, ?string $passphrase = null): self;
```

php/src/Signer.php
```diff
@@ -0,0 +1,111 @@
...
+        $this->jwk = $jwk;
+        $this->kid = $kid;
+        $this->serializer = new CompactSerializer();
+        $this->builder = new JWSBuilder(new AlgorithmManager([ new ES512() ]));
+    }
+
+    /**
+     * @return string
+     * @throws Exceptions\RequestPathNotFoundException
+     * @throws Exception
```
lboynton, line 90: This seem too generic to throw (does it really just throw Exception?)

php/src/Verifier.php
```diff
@@ -0,0 +1,137 @@
...
+    public static function verifyWithKey(JWK $jwk): Verifier
+    {
+        return new self($jwk);
+    }
+
+    /**
+     * @param string $pem
+     * @return Verifier
+     */
+    public static function verifyWithPem(string $pem): Verifier
```
lboynton, line 46: We should probably allow for a passphrase here and on the other verify methods

php/phpunit.xml
```diff
@@ -0,0 +1,18 @@
...
+         colors="true"
+>
+    <testsuites>
+        <testsuite name="Test Suite">
+            <directory suffix="Test.php">./tests</directory>
+        </testsuite>
+    </testsuites>
+    <coverage processUncoveredFiles="true">
+        <include>
+            <directory suffix=".php">./app</directory>
```
lboynton, line 14: Is this a valid dir?

php/src/AbstractJws.php
```diff
@@ -0,0 +1,125 @@
...
+namespace TrueLayer\Signing;
+
+use Exception;
+use TrueLayer\Signing\Contracts\Jws as IJws;
+use TrueLayer\Signing\Exceptions\RequestPathNotFoundException;
+
+abstract class AbstractJws implements IJws
+{
+    /**
+     * @var string
```
lboynton, line 14: I wouldn't change this now, but do we need these type definitions when we have the types defined in the code?

lboynton (approved):

php/phpunit.xml
stefandanaita: Saint Boilerplate ⛪

php/src/AbstractJws.php
stefandanaita: I think PHPStan complains if we don't provide them.

lboynton: I don't remember it complaining about them before :thinking:

alexheretic:
fyi https://github.com/TrueLayer/truelayer-signing/issues/23#issuecomment-984552744

php/src/Contracts/Signer.php
lboynton: Did we decide not make this change?

stefandanaita: I think we've done it?

alexheretic:
fyi new scenario: https://github.com/TrueLayer/truelayer-signing/pull/29

php/src/Contracts/Signer.php
lboynton: No, the params are nullable but they don't default to null currently. So you need to pass in null for the third arg yourself.

>PHP Fatal error:  Uncaught ArgumentCountError: Too few arguments to function TrueLayer\Signing\Signer::signWithPemFile(), 2 passed in test.php on line 9 and exactly 3 expected in TrueLayer/truelayer-signing/php/src/Signer.php:73

PR #17, opened 2021-11-04
C# throw SignatureException for missing jws headers

alexheretic:
C# verify currently throws `System.Collections.Generic.KeyNotFoundException` if required jws header fields are missing, such as `tl_version` or `alg`. This pr instead throws `SignatureException` to handle these simply as invalid signatures.

### DIFF
--- a/csharp/test/ErrorTest.cs
+++ b/csharp/test/ErrorTest.cs
@@ -53,5 +53,24 @@ namespace Tests
 
             verify.Should().Throw<SignatureException>();
         }
+
+         [Fact]
+        public void MissingJwsHeaders()
+        {
+            // jws headers are lacking bits we need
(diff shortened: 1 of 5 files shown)


csharp/test/ErrorTest.cs
```diff
@@ -32,5 +32,24 @@ public void BagSignature()
 
             verify.Should().Throw<SignatureException>();
         }
+
+        [Fact]
```
tl-wajid-malik, line 36: Could you do a Theory with a few cases where a different header is missing each time? Looks like this is testing missing `tl_version` only

alexheretic: Sure I'll take a look tomorrow

tl-wajid-malik (approved):

PR #18, opened 2021-11-05
Kotlin implementation

azanin:
Kotlin/Java implementation to produce & verify TrueLayer API requests signatures.

Kotlin usage:
```kotlin
Signer.from(KID, privateKey)
    .header("Idempotency-Key", idempotencyKey)
    .method("post")
    .path(path)
    .body(body)
    .sign()
```
Java usage:
```java
Signer.from(KID, privateKey)
        .header("Idempotency-Key", idempotencyKey)
        .method("post")
        .path(path)
        .body(body)
        .sign();
```


kotlin/lib/src/main/kotlin/truelayer/signing/Jws.kt
```diff
@@ -0,0 +1,107 @@
...
+
+internal fun verifyTlSignature(
+    signature: String,
+    publicKey: ECPublicKey,
+    requiredHeaders: HashSet<String>,
+    method: String,
+    path: String,
+    body: ByteArray,
+    headers: Map<HeaderName, String>
+): Result<Boolean> {
```
azanin, line 47: Boolean is not a great design choice here, but it is forced by the library we use to verify

.github/workflows/kotlin.yml
```diff
@@ -0,0 +1,27 @@
+name: Kotlin
+
+on:
+  push:
+    branches: [ main, kotlin ]
```
azanin, line 5: remove kotlin branch

kotlin/lib/src/main/kotlin/truelayer/signing/Exceptions.kt
```diff
@@ -0,0 +1,32 @@
...
+
+class InvalidKeyException private constructor(message: String, cause: Throwable) : Exception(message, cause) {
+    companion object {
+        internal fun <T> evaluate(block: () -> T): Result<T> = runCatching {
+            block()
+        }.mapThrowable { e -> InvalidKeyException("Invalid Key: ${e.message}", e) }
+    }
+}
+
+class InvalidSignatureException : Exception {
```
alexheretic, line 11: In C# I called this simply `SignatureException` so could use that here to for "consistency" but its not a big deal.

kotlin/README.md
```diff
@@ -0,0 +1,38 @@
...
+// `Tl-Signature` value to send with the request.
+Signer.from(KID, privateKey)
+    .header("Idempotency-Key", idempotencyKey)
+    .method("post")
+    .path(path)
+    .body(body)
+    .sign()
+```
+
+### Java usage
```
alexheretic, line 15: We'll probably want Java at the top, as it's most likely what clients will be after

kotlin/README.md
```diff
@@ -0,0 +1,38 @@
+# truelayer-signing
+Kotlin/Java package to produce & verify TrueLayer API requests signatures.
+
+### Kotlin usage
+```kotlin
+// `Tl-Signature` value to send with the request.
+Signer.from(KID, privateKey)
```
alexheretic, line 7: We shouldn't imply that `KID` can be const/static when `privateKey` isn't.

```suggestion
Signer.from(kid, privateKey)
```

kotlin/lib/src/main/kotlin/truelayer/signing/Jws.kt
```diff
@@ -0,0 +1,90 @@
...
+
+
+private fun verifyRequiredHeaders(
+    signatureHeaders: JWSHeader,
+    requiredHeaders: HashSet<String>
+): Result<List<String>> {
+    val signatureHeaderNames = signatureHeaders.getCustomParam("tl_headers").toString().split(",").map { it.trim() }
+    return InvalidSignatureException.ensure(
+        { requiredHeaders.all { rHeader -> signatureHeaderNames.any { it.contains(rHeader, true) } } },
+        "missing required header"
```
alexheretic, line 51: best to mention the actual header name that is missing, or at least one of the missing headers names

kotlin/lib/src/main/kotlin/truelayer/signing/Jws.kt
```diff
@@ -0,0 +1,90 @@
...
+    ).map { signatureHeaderNames }
+}
+
+private fun validateSignatureHeaders(signatureHeaders: JWSHeader): Result<Unit> =
+    InvalidSignatureException.ensure(
+        { signatureHeaders.getCustomParam("tl_version").toString() == "2" },
+        "only version 2 is allowed"
```
alexheretic, line 58: This could be confusing, best to mention `tl_version`
```suggestion
        "unsupported jws tl_version"
```

kotlin/lib/src/main/kotlin/truelayer/signing/Jws.kt
```diff
@@ -0,0 +1,90 @@
...
+    )
+
+private fun validateAlghoritm(signatureHeaders: JWSHeader): Result<Unit> =
+    InvalidSignatureException.ensure(
+        { signatureHeaders.algorithm.equals(JWSAlgorithm.ES512) },
+        "unexpected header algorithm"
```
alexheretic, line 64: Saying jws can help avoid confusing between jws headers and http headers that we're also dealing with
```suggestion
        "unsupported jws alg"
```

kotlin/lib/src/main/kotlin/truelayer/signing/Exceptions.kt
alexheretic: The other one similarly could just be `KeyException` if you read "exception" to mean error, which is does, then it is enough.

kotlin/lib/src/main/kotlin/truelayer/signing/Jws.kt
```diff
@@ -0,0 +1,90 @@
...
+            if (value != null) {
+                acc[HeaderName(curr)] = value
+            }
+            acc
+        }
+
+    return InvalidSignatureException.evaluate {
+        JWSObject.parse(signature, Payload(Base64URL(buildPayload(orderedHeaders, method, path, body))))
+            .verify(ECDSAVerifier(publicKey))
+    }.flatMap { verified -> InvalidSignatureException.ensure({ verified }, "jws body mismatch") }
```
alexheretic, line 32: This is misleading, kinda suggests that the http body is different. I'd stick with whats used in other langs at this point `Invalid signature`. The signature is wrong, and thats pretty much all we can say.

```suggestion
    }.flatMap { verified -> InvalidSignatureException.ensure({ verified }, "invalid signature") }
```

kotlin/lib/src/main/kotlin/truelayer/signing/Exceptions.kt
```diff
@@ -0,0 +1,32 @@
...
+    private constructor(message: String, cause: Throwable) : super(message, cause)
+    private constructor(message: String) : super(message)
+
+    companion object {
+        internal fun ensure(predicate: () -> Boolean, message: String): Result<Unit> =
+            if (predicate()) {
+                Result.success(Unit)
+            } else {
+                Result.failure(InvalidSignatureException("JWS signing/verification failed: $message"))
```
alexheretic, line 20: I don't think we need the prefix, it already should be clear from the exception class
```suggestion
                Result.failure(InvalidSignatureException(message))
```

kotlin/lib/src/main/kotlin/truelayer/signing/HeaderName.kt
```diff
@@ -0,0 +1,19 @@
+package truelayer.signing
+
+internal data class HeaderName(val name: String) {
```
alexheretic, line 3: Should add a brief bit of internal docs here to clarify that this implementation is for case-insensitive equals/hash

kotlin/lib/src/main/kotlin/truelayer/signing/Signer.kt
```diff
@@ -0,0 +1,100 @@
...
+    private var method: String = "POST"
+
+    private var path: String = ""
+
+    private var headers: LinkedHashMap<HeaderName, String> = linkedMapOf()
+
+    /**
+     * Add the full request body. Note: This *must* be identical to what is sent with the request.
+     */
+    fun body(body: ByteArray): Signer {
```
alexheretic, line 30: A `String` override for this makes sense I think too

kotlin/lib/src/main/kotlin/truelayer/signing/Verifier.kt
```diff
@@ -0,0 +1,100 @@
...
+     */
+    fun requiredHeader(name: String): Verifier {
+        this.requiredHeaders.add(name)
+        return this
+    }
+
+    /**
+     * Verify the given `Tl-Signature` header value.
+     * @param signature the given `TL-signature`
+     * @return Boolean true id verification succeded or false if unsuccessful
```
alexheretic, line 70: Doesn't return

kotlin/lib/src/main/kotlin/truelayer/signing/Signer.kt
```diff
@@ -0,0 +1,100 @@
...
+    companion object {
+        /**
+         * Start building a request Tl-Signature header value using private key
+         * RFC 7468 PEM-encoded data and the key's kid.
+         * @param kid key identifier of the private key
+         * @param privateKeyPem the privateKey RFC 7468 PEM-encoded
+         * @throws InvalidKeyException if the provided key is invalid
+         */
+        @JvmStatic
+        fun from(kid: String, privateKeyPem: ByteArray): Signer {
```
alexheretic, line 94: Some overrides might make sense too, `String` & pre-parsed `ECPrivateKey` I guess

kotlin/lib/src/main/kotlin/truelayer/signing/Exceptions.kt
```diff
@@ -0,0 +1,32 @@
+package truelayer.signing
+
+class InvalidKeyException private constructor(message: String, cause: Throwable) : Exception(message, cause) {
```
alexheretic, line 3: These are public exceptions so should have some docs too

PR #19, opened 2021-11-13
Java implementation

azanin:
Java implementation to produce & verify TrueLayer API requests signatures.

Java usage:
```java
Signer.from(kid, privateKey)
        .header("Idempotency-Key", idempotencyKey)
        .method("post")
        .path(path)
        .body(body)
        .sign();
```


java/lib/src/main/java/truelayer/signing/Signer.java
```diff
@@ -0,0 +1,187 @@
...
+/**
+ * Builder to generate a Tl-Signature header value using a private key.
+ */
+final public class Signer {
+
+    private final String kid;
+    private final ECPrivateKey ecPrivateKey;
+
+    private String path = "";
+    private String method = "";
```
tl-marco-tormento, line 25: Method should have a default value (e.g. POST).

java/lib/src/main/java/truelayer/signing/Verifier.java
```diff
@@ -0,0 +1,216 @@
...
+import java.util.stream.Stream;
+
+/**
+ * Builder to verify a request against a `Tl-Signature` header using a public key.
+ */
+final public class Verifier {
+
+    private final ECPublicKey publicKey;
+
+    private String method = "";
```
tl-marco-tormento, line 22: Method should have a default value (e.g. POST).

java/lib/src/test/java/truelayer/signing/UsageTest.java
```diff
@@ -0,0 +1,311 @@
...
+                .path(path)
+                .header("X-Whatever", "aoitbeh")
+                .header("Idempotency-Key", idempotencyKey)
+                .body(body)
+                .requiredHeader("Idempotency-Key")
+                .verify(tlSignature); // should not throw
+    }
+
+    @Test
+    public void verifyBodyStaticSignature() throws IOException {
```
alexheretic, line 45: The "body static" probably comes from the rust lib that supports body-only signing too.

```suggestion
    public void verifyStaticSignature() throws IOException {
```

java/lib/src/test/java/truelayer/signing/UsageTest.java
```diff
@@ -0,0 +1,311 @@
...
+                .path(path)
+                .header("X-Whatever-2", "t2345d")
+                .header("Idempotency-Key", idempotencyKey)
+                .body(body)
+                .verify(tlSignature); // should not throw
+    }
+
+
+    @Test
+    public void fullRequestSignatureMethodMismatch() throws IOException {
```
alexheretic, line 65: We can probably remove the `fullRequest` prefix as java won't support v1/body-only signing.

alexheretic (approved):

java/lib/src/main/java/truelayer/signing/Verifier.java
```diff
@@ -0,0 +1,215 @@
...
+
+    private String path = "POST";
```
alexheretic, line 23: wait not this :laughing:

java/lib/src/main/java/truelayer/signing/Verifier.java
alexheretic: Actually I don't think _verifier_ method should have a default, just _sign_.

java/lib/src/main/java/truelayer/signing/Verifier.java
azanin: oh :/ stupid me

alexheretic:
Should update the main README too & add an initial CHANGELOG.md like the other langs.

alexheretic (approved):
:medal_sports:

PR #21, opened 2021-11-16
Removed verifier's method default value and wrote a test for it.

tl-marco-tormento:
As title says, removed verifier's method default value to conform to other implementations.

### DIFF
--- a/go/CHANGELOG.md
+++ b/go/CHANGELOG.md
@@ -4,8 +4,8 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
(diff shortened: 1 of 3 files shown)


go/CHANGELOG.md
```diff
@@ -6,3 +6,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## 0.1.0
 * Added SignWithPem, VerifyWithPem and ExtractJwsHeader.
+
+## 0.1.1
+* Removed verifier's method default value
```
alexheretic, line 11: This should go above, and we should just call it `# Unreleased` unless you want to do a tag for this. I don't mind if you want to tag or not.

tl-marco-tormento: I prefer to tag a new patch version, even if it's a minor thing, for consistency with the other implementations.

alexheretic (approved):

PR #22, opened 2021-11-23
rust: support `jku` jws header

alexheretic:
When signing webhooks we want to also include a `jku` JSON Web Key URL which will be used to lookup the jwk public key to verify webhooks.

This PR supports adding a `jku` when signing & extracting it out of a signature using the rust lib. The former does not need wide support as clients will not need to set `jku` for regular requests. We should probably add extracting the jku to all langs though.

This is a non-breaking change.

### DIFF
--- a/rust/src/jws.rs
+++ b/rust/src/jws.rs
@@ -18,10 +18,17 @@ pub struct JwsHeader {
     /// Comma separated ordered headers used in the signature.
     #[serde(default)]
     pub tl_headers: String,
+    /// JSON Web Key URL. Used in webhook signatures providing the public key jwk url.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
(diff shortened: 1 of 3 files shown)


rust/src/jws.rs
```diff
@@ -18,10 +18,17 @@ pub struct JwsHeader {
     /// Comma separated ordered headers used in the signature.
     #[serde(default)]
     pub tl_headers: String,
+    /// JSON Web Key URL. Used in webhook signatures providing the public key jwk url.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub jku: Option<String>,
```
alexheretic, line 23: This is a non-breaking addition as the struct is marked `#[non_exhaustive]`

tl-marco-tormento (approved):

Issue #23, opened 2021-11-24
Webhook verification support


alexheretic:
Webhook verification is mostly already supported  by the verify implementations. However there are, at least, a couple of ways we can improve the experience for this case. _Note: Actually fetching the jwk should be considered out of scope for the signing libs._

To verify a webhook you need to
```py
# pseudo code
# A
jku = extract_jku(signature)

# B
# fetch jwks (out of scope for signing libs)
jwks = fetch_jwks(jku)

# C
# verify using the jwks
verify_with_jwks(jwks, signature) 
```
So it makes sense that we make **A** & **C** as easy as possible.

* **A** Provide an ability to extract the `jku` from the signature jws header. This is similar to the ability to extract a `kid` and both are needed when verifying a webhook post.
* **C** Provide an easier way to verify using the public key JWK json, instead of requiring callers to lookup kid & convert to pem format themselves. 
It should be possible to provide (using rust as an example) `verify_with_jwks(jwks: &str)` that will lookup the correct jwk by kid & parse etc.

Note: The jwks fetched using the `jku` url will look something like this
```json
{
  "keys": [
    {
      "kty": "RSA",
      "alg": "RS512",
      "kid": "db2e0bb4-e33d-4fc6-a8c5-645e70331127",
      "n": "piC9AfXk1FyhocC77L42j8zWhy-jOeuxdvZCRYACPbU5DB-4zd5JLpTFuYPY5LweqT8tmBbqgXMN_35EQifHrLbBB5a5p-0z3QlVhqMP_K4nu65uIVzC-wjKBcTctqY_jFzSotFZnDNCZEUXcGrTzQifOCy1OD7zT0jm7-1ceWs375joA-KnyJj5rVKBmYrLWa6E2ukLhNBHFzO2N7Aai2w6UFmwfG27viV3_aNkDVC6vkEtZx4oZuCtQ6DKmbiQBTiS1-41wU7QN5PTuA9YUlhOU9yiwnCAG9MDWuJHT7691iLvIIgRctuSQ4HDDg49M39Un4DBt51KCoZsAwxtNw",
      "e": "AQAB"
    },
    {
      "kty": "EC",
      "alg": "EC",
      "kid": "07035ce6-c8a8-42fc-b38b-d843d4c7d40e",
      "crv": "P-521",
[...]

PR #27, opened 2021-12-02
Golang add VerifyWithJwks and jku extraction

tl-marco-tormento:
Adds method to verify using the `jwks` response data directly.
Also adds a `jku` field to `jws` header to allow extracting webhook signature `jku`.

See #23 for motivating details.

```go
jwsHeader, err := ExtractJwsHeader(webhookSignature)
assert.Equal(jwsHeader.Jku, "https://webhooks.truelayer.com/.well-known/jwks")
```

```go
err := VerifyWithJwks(jwks).
    Method("POST").
    Path("/tl-webhook").
    Header("x-tl-webhook-timestamp", []byte("2021-11-29T11:42:55Z")).
    Header("content-type", []byte("application/json")).
    Body([]byte("{\"event_type\":\"example\",\"event_id\":\"18b2842b-a57b-4887-a0a6-d3c7c36f1020\"}")).
    Verify(webhookSignature)
```


go/crypto/crypto.go
```diff
@@ -96,3 +117,49 @@ func VerifyES512(key *ecdsa.PublicKey, payload []byte, signature []byte) error {
...
+}
+
+type Jwk struct {
+	Kid string `json:"kid"`
+	Kty string `json:"kty"`
+	Alg string `json:"alg"`
+	Crv string `json:"crv"`
+	X   string `json:"x"`
+	Y   string `json:"y"`
+}
```
conradludgate, line 133: Do we want these types to be public? It seems they're only used internally

conradludgate (approved):

alexheretic (approved):

PR #35, opened 2022-01-28
Add verifying webhooks readme examples for golang

tl-marco-tormento:
Adds golang verify webhook snippet examples to their READMEs.

### DIFF
--- a/go/CHANGELOG.md
+++ b/go/CHANGELOG.md
@@ -4,6 +4,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+* Added verifying webhooks readme examples
+
 ## 0.1.2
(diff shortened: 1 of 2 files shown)


go/CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+* Added verifying webhooks readme examples
+
```
alexheretic, line 9: Is updating the readme a "change"?

alexheretic (approved):

go/CHANGELOG.md
tl-marco-tormento: I'm not sure tbh.
Is documentation the `product` or just something that goes with it as some sort of metadata?
In the former case updating it is a "change", in the latter it's not.

tl-marco-tormento: Since this is a public library I reckon documentation is an important piece of the product?

PR #37, opened 2022-02-16
Support jwk.alg being "ES512"

alexheretic:
Referring to https://datatracker.ietf.org/doc/html/draft-ietf-jose-json-web-algorithms#section-3.1 our EC jwk.alg field should actually be `ES512` rather than `EC`.

This pr adds support for that without breaking the existing alg value, since this field isn't strictly required anyway.

Affects langs: **rust, go, java, csharp**


alexheretic:
We should update the changlogs too...

tl-marco-tormento (approved):

azanin (approved):

PR #39, opened 2022-02-24
Add `headers` method to Signer and Verifier

azanin:
- Add `headers` method to `Signer` and `Verifier`
- Update README
- Remove scala and kotlin examples


java/lib/src/main/java/truelayer/signing/Verifier.java
```diff
@@ -105,6 +105,23 @@ public Verifier body(String body) {
...
 
+    /**
+     * Add a Map of headers.
+     * Warning: Only a single value per header name is supported.
+     *
+     * @param headers  - must not be null
+     * @return the Verifier instance
+     * @throws IllegalArgumentException if the provided params are null
+     */
+    public Verifier headers(Map<String, String> headers) {
```
alexheretic, line 116: `Map` makes sense to have :+1: Are there other obvious std-lib types that http libraries would use for headers? We could add overloads if so...

alexheretic (approved):

java/lib/src/main/java/truelayer/signing/Verifier.java
azanin: we can add later if needed

PR #42, opened 2022-03-08
Move under com.truelayer and define release pipeline

azanin:
This PR adds:
- Release workflow. Push to maven central only if the `gradle.properties` version attribute changed from the previous commit. It will just close the staging repository at first, so promotion will be manual.
- Move package `truelayer.signing` to `com.truelayer.signing`

When merged it will not trigger a release.
I will do another PR with Changelog changes and version set to `0.2.0`.

### DIFF
--- a/.github/workflows/java.yml
+++ b/.github/workflows/java.yml
@@ -2,26 +2,98 @@ name: Java
 
 on:
   push:
-    branches: [ main ]
-  pull_request:
-    branches: [ main ]
+    branches:
+      - '**'
+    tags-ignore:
+      - '**'
 
 jobs:
-  test:
-    name: Test
+  build-and-test:
+    name: Build and Test
     strategy:
+      fail-fast: false
       matrix:
         os: [ ubuntu-latest ]
-        java: [ 8, 11, 17 ]
+        java: [ 8, 9, 10, 11, 12, 13, 14, 15, 16, 17 ]
     runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout current branch
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       - name: Setup JDK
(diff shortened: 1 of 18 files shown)


.github/workflows/java.yml
```diff
@@ -2,26 +2,98 @@ name: Java
...
-  test:
-    name: Test
+  build-and-test:
+    name: Build and Test
     strategy:
+      fail-fast: false
       matrix:
         os: [ ubuntu-latest ]
-        java: [ 8, 11, 17 ]
+        java: [ 8, 9, 10, 11, 12, 13, 14, 15, 16, 17 ]
```
tl-andrea-dilisio, line 17: it would be nice to define a consistent approach between this library and the Java SDK. Currently we don't have a matrix on the library but we can onboard it

.github/workflows/java.yml
```diff
@@ -2,26 +2,98 @@ name: Java
...
+    runs-on: ubuntu-latest
+    if: github.ref == 'refs/heads/main'
+    steps:
+      - name: Checkout current branch
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 2 #we need to make a git diff of the last 2 commits
+      - name: Get project version
+        id: get_project_version
+        run: |
```
tl-andrea-dilisio, line 44: As discussed, this is a different approach compared to the Java SDK but we agreed that's fine having projects-tailored approaches that diverge a bit

tl-andrea-dilisio (approved):

.github/workflows/java.yml
azanin: I thought it could be useful. I just based this on other libraries' approaches such as [retrofit](https://github.com/square/retrofit/blob/master/.github/workflows/build.yml#L12)

PR #44, opened 2022-03-08
Update php verification headers

tl-ozum-safaoglu:


alexheretic:
The problem with this is PHP is documenting the server-side style verification code. Other langs don't include this in the README because it isn't a client facing function. It exists server side and to guarantee signing algorithm consistency cross language.

So we should probably just remove the verify snippet from the readme and properly develop a jwks verify functionality for PHP as detailed in https://github.com/TrueLayer/truelayer-signing/issues/23

alexheretic:
Is this pr still relevant @tl-ozum-safaoglu ?

PR #50, opened 2022-03-21
[PAYG-715] Fix C# parsing jwks with uneven EC coord byte lengths

alexheretic:
The current test-resources jwks has coords both 66 bytes long. However, in some cases EC coords may be generated that don't use the most significant bit(s) and so may be encoded as shorter byte array representations.

Some JWK parsing implementations may have trouble with this, e.g. C# _(Update: it was just C#)_
> System.Security.Cryptography.CryptographicException : The specified key parameters are not valid. Q.X and Q.Y, or D, must be specified. Q.X, Q.Y must be the same length. If D is specified it must be the same length as Q.X and Q.Y if also specified for named curves or the same length as Order for explicit curves.

## Reproduction
To surface this issue I've regenerated the EC keys in ./test-resources so that the JWK will have x=65 bytes, y=66 bytes.

See https://github.com/TrueLayer/truelayer-signing/runs/5630830983?check_suite_focus=true#step:5:38

## Fix
For such implementations the coord byte representations should be prepend zero padded (as they're big endian) to 66 bytes.


csharp/src/Verifier.cs
```diff
@@ -238,8 +238,10 @@ private void FindAndImportJwk(Jwks jwks, string kid)
                 Curve = ECCurve.CreateFromFriendlyName("secp521r1"),
                 Q = new ECPoint
                 {
-                    X = Base64Url.Decode(jwk.X),
-                    Y = Base64Url.Decode(jwk.Y),
+                    // Note: A CryptographicException will be thrown if the coord byte
+                    // representations have uneven length, so must be zero padded to 66
+                    X = Base64Url.Decode(jwk.X).PrependZeroPad(66),
+                    Y = Base64Url.Decode(jwk.Y).PrependZeroPad(66),
```
alexheretic, line 244: Note: This was the fix, no other langs were affected.

tl-wajid-malik (approved):

Issue #51, opened 2022-03-23
Provide working request signing examples against `/test-signature` endpoint


alexheretic:
It would be helpful to have runnable examples in each lang `POST https://api.truelayer-sandbox.com/test-signature`. 

**Example readme template:**
---

# $LANG request signature example
Sends a signed request to `https://api.truelayer-sandbox.com/test-signature`.

## Run
Set environment variables:
* `ACCESS_TOKEN` A valid JWT access token for `payments` scope [docs](https://docs.truelayer.com/docs/retrieve-a-token-in-your-server).
* `KID` The certificate/key UUID for associated with your public key uploaded to console.truelayer.com.
* `PRIVATE_KEY` Private key PEM string that matches the `KID` & uploaded public key.
  Should have the same format as [this example private key](https://github.com/TrueLayer/truelayer-signing/blob/main/test-resources/ec512-private.pem).

```sh
$ $command-to-run-example
204 ✓
```

alexheretic:
Java taken care of in #56

PR #54, opened 2022-03-24
[PAYG-723] Adds rust tl-signature example

tl-flavio-barinas:
#51 for Rust

# Rust request signature example
Sends a signed request to `https://api.truelayer-sandbox.com/test-signature`.

## Run
Set environment variables:
* `ACCESS_TOKEN` A valid JWT access token for `payments` scope [docs](https://docs.truelayer.com/docs/retrieve-a-token-in-your-server).
* `KID` The certificate/key UUID for associated with your public key uploaded to console.truelayer.com.
* `PRIVATE_KEY` Private key PEM string that matches the `KID` & uploaded public key.
  Should have the same format as [this example private key](https://github.com/TrueLayer/truelayer-signing/blob/main/test-resources/ec512-private.pem).

```sh
$ cargo run
204 No Content ✓
```

### DIFF
--- a/rust/examples/sign-request/src/main.rs
+++ b/rust/examples/sign-request/src/main.rs
@@ -1,0 +1,50 @@
+use rand;
+use std::env;
+use uuid::Uuid;
+
+const KID: &str = "KID";
+const ACCESS_TOKEN: &str = "ACCESS_TOKEN";
+const PRIVATE_KEY: &str = "PRIVATE_KEY";
+
+const URL: &str = "https://api.truelayer-sandbox.com/test-signature";
+
+#[tokio::main]
+async fn main() {
+    // load env vars
+    let kid = env::var(KID).expect("Missing env var KID");
+    let access_token = env::var(ACCESS_TOKEN).expect("Missing env var ACCESS_TOKEN");
+    let private_key = env::var(PRIVATE_KEY).expect("Missing env var PRIVATE_KEY");
+
+    // create idemoptency key and body
+    let idempotency_key = Uuid::new_v4().to_string();
+    let body = format!("body-{}", rand::random::<u32>());
+
+    // generate tl-signature
+    let tl_signature = truelayer_signing::sign_with_pem(kid.as_str(), private_key.as_bytes())
+        .method("POST")
+        .path("/test-signature")
+        .header("Idempotency-Key", idempotency_key.as_bytes())
+        .header("X-Bar-Header", b"abc123")
+        .body(body.as_bytes())
+        .sign()
+        .unwrap();
+
+    // call `/test-signature` endpoint
+    let client = reqwest::Client::new();
+    let response = client
+        .post(URL)
+        .header("Authorization", format!("Bearer {access_token}"))
+        .header("Idempotency-Key", idempotency_key)
+        .header("X-Bar-Header", "abc123")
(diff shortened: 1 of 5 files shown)


rust/examples/sign-request/src/main.rs
```diff
@@ -0,0 +1,50 @@
...
+        .header("Tl-Signature", tl_signature)
+        .body(body)
+        .send()
+        .await
+        .unwrap();
+
+    // print result
+    match response.error_for_status_ref() {
+        Ok(res) => println!("{} ✓", res.status()),
+        Err(err) => println!("{}", err.status().unwrap()),
```
alexheretic, line 48: It's helpful to print the response body when it fails.

rust/examples/sign-request/src/main.rs
```diff
@@ -0,0 +1,50 @@
...
+const PRIVATE_KEY: &str = "PRIVATE_KEY";
+
+const URL: &str = "https://api.truelayer-sandbox.com/test-signature";
+
+#[tokio::main]
+async fn main() {
+    // load env vars
+    let kid = env::var(KID).expect("Missing env var KID");
+    let access_token = env::var(ACCESS_TOKEN).expect("Missing env var ACCESS_TOKEN");
+    let private_key = env::var(PRIVATE_KEY).expect("Missing env var PRIVATE_KEY");
```
alexheretic, line 16: I think just inline strings would be better here
```suggestion
    let kid = env::var("KID").expect("Missing env var KID");
    let access_token = env::var("ACCESS_TOKEN").expect("Missing env var ACCESS_TOKEN");
    let private_key = env::var("PRIVATE_KEY").expect("Missing env var PRIVATE_KEY");
```

rust/examples/sign-request/src/main.rs
```diff
@@ -0,0 +1,50 @@
...
+        .header("Idempotency-Key", idempotency_key.as_bytes())
+        .header("X-Bar-Header", b"abc123")
+        .body(body.as_bytes())
+        .sign()
+        .unwrap();
+
+    // call `/test-signature` endpoint
+    let client = reqwest::Client::new();
+    let response = client
+        .post(URL)
```
alexheretic, line 35: I think this is better inline too, so you can see it near the `.path(...)` declaration. You wouldn't be able to change the const easily to another endpoint without being aware of that anyway.

alexheretic:
Lets add a link in the README too, like https://github.com/TrueLayer/truelayer-signing/blob/main/nodejs/README.md?plain=1#L25

rust/examples/sign-request/src/main.rs
tl-flavio-barinas: done on eafddd496b69dbb8c00a64c677ccae7e80d95761

alexheretic:
Great! It would be good to have similar comments to what we have in the nodejs & C# ones too.

E.g. explaining each bit of the signing call and reiterating
```
// Request body & any signed headers *must* exactly match what was used to generate the signature.
```

Probably just worth copying those comments into the rust example in the right places.

alexheretic (approved):
great stuff :tada:

alexheretic:
I'm not sure why only the java actions are running though, this happened on the recent C# pr too..but not the nodejs one :thinking:

Issue #47, opened 2022-03-11
Provide webhook verifying server examples


alexheretic:
It would be helpful to see example server code receiving and verifying webhooks against TrueLayer's public jwks.

alexheretic:
Java done in #56

PR #58, opened 2022-04-11
[PHP-32] Add support for signature verification using JWKS json

stefandanaita:


lboynton:
Changelog?

lboynton:
Does this address #44?

alexheretic:
Yeah as I said in https://github.com/TrueLayer/truelayer-signing/pull/44#issuecomment-1063140805 I think the php readme could be a little more consistent with the other langs. In particular, we don't need to mention server-side verification which we don't expect clients to use. Instead a simple signing snippet & a jwks webhook verify snippet shows the two ways we expect clients to use this.

And it would great to see #47 & #51 examples later too!

stefandanaita:
> Yeah as I said in [#44 (comment)](https://github.com/TrueLayer/truelayer-signing/pull/44#issuecomment-1063140805) I think the php readme could be a little more consistent with the other langs. In particular, we don't need to mention server-side verification which we don't expect clients to use. Instead a simple signing snippet & a jwks webhook verify snippet shows the two ways we expect clients to use this.
> 
> And it would great to see #47 & #51 examples later too!

Sorry, I was on holiday for the past weeks. I'll take care of those points separately. For now we need this merged in so that we can do jwks fetching & signature verification in the client library.

lboynton (approved):

PR #61, opened 2022-04-13
[MAP-101] Add Golang request signature example

tl-marco-tormento:
#51 for Golang

# Golang request signature example
Sends a signed request to `https://api.truelayer-sandbox.com/test-signature`.

## Run
Set environment variables:
* `ACCESS_TOKEN` A valid JWT access token for `payments` scope [docs](https://docs.truelayer.com/docs/retrieve-a-token-in-your-server).
* `KID` The certificate/key UUID for associated with your public key uploaded to console.truelayer.com.
* `PRIVATE_KEY` Private key PEM string that matches the `KID` & uploaded public key.
  Should have the same format as [this example private key](https://github.com/TrueLayer/truelayer-signing/blob/main/test-resources/ec512-private.pem).

```sh
$ go run main.go
Sending...
204 ✓
```

### DIFF
--- a/go/examples/sign-request/main.go
+++ b/go/examples/sign-request/main.go
@@ -52,9 +52,10 @@ func main() {
 		Body([]byte(body)). // body of our request
 		Sign()
 
-	client := &http.Client{}
+	fmt.Println("Sending...")
 
 	// Request body & any signed headers *must* exactly match what was used to generate the signature.
+	client := &http.Client{}
 	req, err := http.NewRequest("POST", fmt.Sprintf("%s/test-signature", TlBaseUrl), strings.NewReader(body))
 	req.Header.Add("Authorization", fmt.Sprintf("Bearer %s", accessToken))
 	req.Header.Add("Idempotency-Key", idempotencyKey)
@@ -81,5 +82,5 @@ func main() {
 
 	// 204 means success
 	// 401 means either the access token is invalid, or the signature is invalid.
-	fmt.Printf("%d %s", statusCode, responseBody)
+	fmt.Printf("%d %s\n", statusCode, responseBody)
 }
--- a/go/examples/sign-request/main.go
+++ b/go/examples/sign-request/main.go
@@ -14,7 +14,7 @@ import (
 
 // the base url to use
 const (
-	TlBaseUrl = "https://api.truelayer-sandbox.com"
+	TlBaseURL = "https://api.truelayer-sandbox.com"
 )
 
 func main() {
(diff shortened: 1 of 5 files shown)


alexheretic (approved):
lgtm

```
$ go run main.go
Sending...
204 ✓
```

azanin (approved):

conradludgate (approved):

go/examples/sign-request/main.go
```diff
@@ -0,0 +1,86 @@
...
+			responseBody = []byte("✓")
+		} else {
+			defer resp.Body.Close()
+			responseBody, err = io.ReadAll(resp.Body)
+			if err != nil {
+				responseBody = []byte(fmt.Sprintf("%s", err.Error()))
+			}
+		}
+	} else {
+		responseBody = []byte(fmt.Sprintf("%s", err.Error()))
```
conradludgate, line 80: Why are you formatting these?

go/examples/sign-request/main.go
```diff
@@ -0,0 +1,86 @@
...
+	"os"
+	"strings"
+
+	tlsigning "github.com/Truelayer/truelayer-signing/go"
+	"github.com/google/uuid"
+)
+
+// the base url to use
+const (
+	TlBaseUrl = "https://api.truelayer-sandbox.com"
```
conradludgate, line 17: Go usually uses full-upper acronyms if I remember correctly
```suggestion
	TLBaseURL = "https://api.truelayer-sandbox.com"
```

go/examples/sign-request/main.go
```diff
@@ -0,0 +1,86 @@
...
+		Header("Idempotency-Key", []byte(idempotencyKey)).
+		Header("X-Bar-Header", []byte("abc123")).
+		Body([]byte(body)). // body of our request
+		Sign()
+
+	fmt.Println("Sending...")
+
+	// Request body & any signed headers *must* exactly match what was used to generate the signature.
+	client := &http.Client{}
+	req, err := http.NewRequest("POST", fmt.Sprintf("%s/test-signature", TlBaseUrl), strings.NewReader(body))
```
conradludgate, line 59: Unchecked error ;)

go/examples/sign-request/main.go
skos-ninja: Yeah in this case `.Error()` is a string so shouldn't need to be formatted. There's also should we be converting the strings to bytes and then back to strings again or just converting the response body to string and then using that below

go/examples/sign-request/main.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	// A random body string is enough for this request as `/test-signature` endpoint does not
+	// require any schema, it simply checks the signature is valid against what's received.
+	body := fmt.Sprintf("body-%d", rand.Intn(99999999))
+
+	idempotencyKey := uuid.New().String()
+
+	// Generate tl-signature
+	tlSignature, err := tlsigning.SignWithPem(kid, []byte(privateKey)).
```
skos-ninja, line 45: Unchecked error here too

tl-marco-tormento: Our signing library never fails. 😎

PR #62, opened 2022-04-13
[MAP-104] add python tl-signature lib

tl-flavio-barinas:


python/examples/webhook-server/main.py
```diff
@@ -0,0 +1,70 @@
...
+class HookHandler(BaseHTTPRequestHandler):
+    def do_POST(self):
+        if self.path == HOOK_PATH:
+            content_length = int(self.headers['Content-Length'])
+            body = self.rfile.read(content_length)
+
+            try:
+                res = hook_handler(self.path, self.headers, body.decode())
+                self.send_response(res)
+            except Exception as err:
```
tl-rodrigo-gryzinski, line 22: I'd make TL verification functions return a specific exception type (or exceptions deriving from a common base) so that we could return `UNAUTHORIZED` only for actual signature validation failures, not errors in `send_response` or or other errors

python/truelayer_signing/sign.py
```diff
@@ -0,0 +1,82 @@
...
+            self.http_method
+        ))
+
+
+@dataclass(frozen=True)
+class SignArguments:
+    kid: str
+    pkey: str
+    path: str
+    headers: Mapping[str, str]
```
tl-rodrigo-gryzinski, line 51: Ordering is important here, since we're iterating over the keys to create `tl_headers` the user will need to use the same instance of headers when making the request (if using a `dict` as the `Mapping` instance) since the keys are not guaranteed to have the same ordering across different `dict` instances.

I think making this an `OrderedDict` would help making this explicit in the API though then users would be burdened with not being able to use a simple `dict`, alternatively we can add a notice in the docs/examples to make this very clear

python/truelayer_signing/utils.py
```diff
@@ -0,0 +1,143 @@
...
+    PATCH = "PATCH"
+    PUT = "PUT"
+    DELETE = "DELETE"
+
+
+class TlJwsBase:
+    pkey: str
+    http_method: HttpMethod
+    path: str
+    headers: MutableMapping[str, str]
```
tl-rodrigo-gryzinski, line 24: Same caveat as in `sign.py` applies here, ordering matters and `dict`s have an implicit and unspecified order that changes from instance to instance

python/truelayer_signing/utils.py
```diff
@@ -0,0 +1,143 @@
...
+
+    def set_path(self: T, path: str) -> T:
+        """
+        Add the request absolute path starting with a leading `/` and without
+        any trailing slashes.
+        """
+        self.path = path
+        return self
+
+    def add_headers(self: T, headers: MutableMapping[str, str]) -> T:
```
tl-rodrigo-gryzinski, line 56: Since ordering matters, what do you think about making headers an `Iterable[Tuple[str, str]]` or `Iterator[...]`? Users would have to call `.items()` when passing `dict`s but then at least they're making a concious choice of relying on the unspecified ordering of `dict` keys

python/truelayer_signing/utils.py
```diff
@@ -0,0 +1,143 @@
...
+    ```
+    """
+    payload = f"{method} {path}\n"
+    for key, value in headers.items():
+        payload += f"{key}: {value}\n"
+    payload += body
+    return payload
+
+
+def build_v2_jws_b64(
```
tl-rodrigo-gryzinski, line 110: We could catch and return exceptions wrapped in our tl exception type for json encoding errors. If I were using this API I would like to be sure that catching `TlSigningException` would handle expected errors like json and base64 encoding errors, what do you think?

python/truelayer_signing/verify.py
```diff
@@ -0,0 +1,132 @@
...
+    """
+    key_fmt: KeyFmt
+
+    def __init__(
+        self,
+        pkey: str,
+        key_fmt: KeyFmt,
+        method: HttpMethod = HttpMethod.POST,
+        path: str = "",
+        headers: Optional[Dict[str, str]] = None,
```
tl-rodrigo-gryzinski, line 33: Same notes about `dict` ordering

python/truelayer_signing/verify.py
```diff
@@ -0,0 +1,132 @@
...
+    tl_signature: str
+    pkey: str
+    key_fmt: KeyFmt
+    path: str
+    headers: Mapping[str, str]
+    body: str
+    method: HttpMethod
+
+
+def tl_verify(args: VerifyArguments) -> bool:
```
tl-rodrigo-gryzinski, line 65: I'd raise specific exception types from here, looking at the webhook example it'd be nice for users to be able to use some catch-all verification exception to return `UNAUTHORIZED` or some other specific way of handling this

python/truelayer_signing/utils.py
tl-rodrigo-gryzinski, line 56: Wouldn't just `Mapping` be sufficient here?

tl-flavio-barinas: Updated to just be `Mapping`

python/truelayer_signing/utils.py
tl-flavio-barinas: We should never get a JSON error unless the function is used improperly as this is an internal function I think it's a reasonable assumetion. I did add some documentation that a `UnicodeEncodeError` can be raised. 

What do you think? would it be worth wrapping the exception

python/truelayer_signing/verify.py
tl-flavio-barinas: Updated to use a custom exception. Let me know what you think.

python/truelayer_signing/utils.py
alexheretic: iirc in other libs we wrapped all such errors, it makes it easier for users verifying a signature to handle a single exception for any expected case, e.g. sending some nonsense signature header.

python/README.md
```diff
@@ -0,0 +1,36 @@
+# truelayer-signing
+Python package to produce & verify TrueLayer API requests signatures.
+
+```
+tl_signature = sign_with_pem(KID, PRIVATE_KEY) \
+    .set_method(HttpMethod.POST) \
+    .set_path(path) \
+    .add_header("Idempotency-Key", idempotency_key) \
+    .set_body(body) \
+    .sign()
```
alexheretic, line 10: One thing I was wondering about python is if this builder is idiomatic? My experience with python you tend to have a single method call with multiple (maybe optional) arguments. Wdyt?

For similar reasons I used a different API style for nodejs: https://github.com/TrueLayer/truelayer-signing/tree/main/nodejs#usage

tl-rodrigo-gryzinski: I think taking a big list of parameters is more common in python due to named arguments but I wouldn't say the builder is not idiomatic, python code varies too much for there to be any standards in my opinion. I do lean slightly in favour of a single method

tl-rodrigo-gryzinski (approved):
LGTM, though there's the open discussion about the builder

python/truelayer_signing/utils.py
tl-flavio-barinas: Good point, I hadn't considered the verification route. Will set a custom error.

python/README.md
alexheretic: Ok as long as we have some angle to call it idiomatic then I'm ok with the builder style.

alexheretic:
I'm having some issues getting building/running the code with poetry. (I'm getting `ImportError: cannot import name 'Tag' from partially initialized module 'packaging.tags' (most likely due to a circular import) (/home/alex/.poetry/lib/poetry/_vendor/py3.10/poetry/core/_vendor/packaging/tags.py)`.)

But if this is working fine for others then probably ok to merge. I'll try to fix my setup next week and verify the examples etc.

alexheretic:
We should add a python link to the top-level readme too.

alexheretic:
> I'm having some issues getting building/running the code with poetry....

Ok I manage to get poetry to work :) The examples aren't trivial to run though, they don't use poetry. Why don't we add `requests` as a dev-dependency and let these be run that way?

alexheretic:
Btw both examples work fine for me :+1:

python/tests/test_tl_sign.py
```diff
@@ -0,0 +1,82 @@
...
+
+    with pytest.raises(Exception):
+        verify_with_pem(PUBLIC_KEY) \
+            .set_method(HttpMethod.POST) \
+            .set_path("/foo") \
+            .set_body("{}") \
+            .verify(tl_signature)
+
+
+def mismatched_signature_with_attached_valid_body_trailing_dots():
```
alexheretic, line 68: This test doesn't run, I think it needs `test_` in front of it.

python/tests/test_tl_sign.py
```diff
@@ -0,0 +1,82 @@
...
+    # signature for `/bar` but with a valid jws-body pre-attached
+    # if we run a simple jws verify on this unchanged it'll work!
+    tl_signature = "eyJhbGciOiJFUzUxMiIsImtpZCI6IjQ1ZmM3NWNmLTU2ND\
+      ktNDEzNC04NGIzLTE5MmMyYzc4ZTk5MCIsInRsX3ZlcnNpb24iOiIyIiwidGxfaGV\
+      hZGVycyI6IiJ9.UE9TVCAvYmFyCnt9.ARLa7Q5b8k5CIhfy1qrS-IkNqCDeE-VFRD\
+      z7Lb0fXUMOi_Ktck-R7BHDMXFDzbI5TyaxIo5TGHZV_cs0fg96dlSxAERp3UaN2oC\
+      QHIE5gQ4m5uU3ee69XfwwU_RpEIMFypycxwq1HOf4LzTLXqP_CDT8DdyX8oTwYdUB\
+      d2d3D17Wd9UA"
+
+    with pytest.raises(Exception):
```
alexheretic, line 60: This should raise `TlSigningException` right?
```suggestion
    with pytest.raises(TlSigningException):
```

python/tests/test_tl_sign.py
```diff
@@ -0,0 +1,82 @@
+import pytest
```
alexheretic, line 1: Seems to be missing a few scenarios
* signature_no_headers
* signature_method_mismatch
* signature_path_mismatch
* signature_header_mismatch
* signature_body_mismatch
* signature_missing_signature_header
* signature_required_header_missing_from_signature
* flexible_header_case_order_verify
* extract_kid / extract_jku (or whole jws header however that's done here)
* verify_with_jwks

python/examples/webhook-server/main.py
```diff
@@ -0,0 +1,89 @@
...
+        raise ValueError(f"Unpermitted jku {jku}")
+
+    # fetch jkws
+    res = requests.get(jku)
+    res.raise_for_status()
+    keys = res.json()["keys"]
+    try:
+        jwks = next(filter(lambda x: x["kid"] == jws_header["kid"], keys))
+    except StopIteration:
+        raise ValueError("no jwk found for signature kid")
```
tl-flavio-barinas, line 58: should this logic be part of the library?

python/truelayer_signing/verify.py
```diff
@@ -0,0 +1,227 @@
...
+    Tl-Verifier
+    """
+    key_fmt: KeyFmt
+    required_headers: List[str]
+
+    def __init__(
+        self,
+        pkey: Union[str, Mapping[str, str]],
+        key_fmt: KeyFmt,
+        method: HttpMethod = HttpMethod.POST,
```
alexheretic, line 42: We don't default this in the verifier in other languages, so maybe best to be consistent & require it be specified here too.

python/tests/test_tl_sign.py
```diff
@@ -0,0 +1,291 @@
...
+        .add_header("idempotency-key", idempotency_key) \
+        .set_body(body) \
+        .verify(tl_signature)
+
+
+def test_extract_jws_header():
+    hook_signature = read_file("../test-resources/webhook-signature.txt")
+    jws_header = extract_jws_header(hook_signature)
+    assert(jws_header["alg"] == "ES512")
+    assert(jws_header["alg"] == "ES512")
```
alexheretic, line 263:
```suggestion
    assert(jws_header["alg"] == "ES512")
```

alexheretic (approved):
Great work :medal_sports:

PR #63, opened 2022-04-13
[MAP-24] Add Golang webhook server example

tl-marco-tormento:
#47 for Golang, including jwks fetch caching.

# Golang webhook server example
A http server than can receive and verify signed TrueLayer webhooks.

## Run
Run the server.
```sh
go run main.go
```

Send a valid webhook that was signed for path `/hook/d7a2c49d-110a-4ed2-a07d-8fdb3ea6424b`.
```sh
curl -iX POST -H "Content-Type: application/json" \
    -H "X-Tl-Webhook-Timestamp: 2022-03-11T14:00:33Z" \
    -H "Tl-Signature: eyJhbGciOiJFUzUxMiIsImtpZCI6IjFmYzBlNTlmLWIzMzUtNDdjYS05OWE5LTczNzQ5NTc1NmE1OCIsInRsX3ZlcnNpb24iOiIyIiwidGxfaGVhZGVycyI6IngtdGwtd2ViaG9vay10aW1lc3RhbXAiLCJqa3UiOiJodHRwczovL3dlYmhvb2tzLnRydWVsYXllci5jb20vLndlbGwta25vd24vandrcyJ9..AE_QsBRhnsMkcRzd42wvY1e2HruUhkOgjuZKktGH_WmbD7rBzoaEHUuF36IxyyvCbLajd3MBExNmzjbrOQsGaspwAI5DcGVMFLKUhB7ZzUlTP9up3eNUrdwWyyfBWDQb-qmEuLnrhFDJvgCXEqlV5OLrt-O7LaRAJ4f9KHsZLQ_j2vPC" \
    -d "{\"event_type\":\"payout_settled\",\"event_schema_version\":1,\"event_id\":\"8fb9fb4e-bb2b-400b-af64-59e5dde74bad\",\"event_body\":{\"transaction_id\":\"c34c8721-66a9-49f6-a229-284efcf88a02\",\"settled_at\":\"2022-03-11T14:00:32.933000Z\"}}" \
    http://localhost:7000/hook/d7a2c49d-110a-4ed2-a07d-8fdb3ea6424b
```

Modifying the `X-Tl-Webhook-Timestamp` header, the body or the path will cause the above signature to be invalid.

### DIFF
--- a/go/examples/webhook-server/go.mod
+++ b/go/examples/webhook-server/go.mod
@@ -1,0 +1,9 @@
+module webhook-server
+
+go 1.18
+
+require (
+	github.com/Truelayer/truelayer-signing/go v0.1.4 // indirect
+	github.com/gregjones/httpcache v0.0.0-20190611155906-901d90724c79 // indirect
+	github.com/wk8/go-ordered-map v0.2.0 // indirect
+)
--- a/go/examples/webhook-server/main.go
+++ b/go/examples/webhook-server/main.go
@@ -11,73 +11,75 @@ import (
 )
 
 func main() {
-	http.HandleFunc("/hook/d7a2c49d-110a-4ed2-a07d-8fdb3ea6424b", handler)
+	tp := httpcache.NewMemoryCacheTransport()
+	client := http.Client{Transport: tp}
+
+	http.HandleFunc("/hook/d7a2c49d-110a-4ed2-a07d-8fdb3ea6424b", handler(&client))
 
 	log.Println("Starting server on: 7000")
 
 	log.Fatal(http.ListenAndServe(":7000", nil))
 }
 
-func handler(w http.ResponseWriter, r *http.Request) {
-	if r.Method != "POST" {
-		http.Error(w, "Method is not supported.", http.StatusNotFound)
-		return
-	}
+func handler(client *http.Client) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		if r.Method != http.MethodPost {
+			http.Error(w, "Method is not supported.", http.StatusNotFound)
+			return
+		}
 
-	verified, err := verifyHook(r)
-	if verified {
-		w.WriteHeader(http.StatusAccepted)
-	} else {
-		log.Println(err)
-		w.WriteHeader(http.StatusUnauthorized)
+		err := verifyHook(client, r)
+		if err == nil {
+			w.WriteHeader(http.StatusAccepted)
+		} else {
+			log.Println(err)
+			w.WriteHeader(http.StatusUnauthorized)
+		}
 	}
 }
 
-func verifyHook(r *http.Request) (bool, error) {
+func verifyHook(client *http.Client, r *http.Request) error {
 	tlSignature := r.Header.Get("Tl-Signature")
 	if len(tlSignature) == 0 {
-		return false, fmt.Errorf("missing Tl-Signature header")
+		return fmt.Errorf("missing Tl-Signature header")
 	}
 
 	jwsHeader, err := tlsigning.ExtractJwsHeader(tlSignature)
 	if err != nil {
-		return false, fmt.Errorf("jku missing")
+		return fmt.Errorf("jku missing")
 	}
 
 	defer r.Body.Close()
 	webhookBody, err := io.ReadAll(r.Body)
 	if err != nil {
-		return false, fmt.Errorf("webhook body missing")
+		return fmt.Errorf("webhook body missing")
 	}
 
 	// ensure jku is an expected TrueLayer url
 	if jwsHeader.Jku != "https://webhooks.truelayer.com/.well-known/jwks" && jwsHeader.Jku != "https://webhooks.truelayer-sandbox.com/.well-known/jwks" {
-		return false, fmt.Errorf("unpermitted jku %s", jwsHeader.Jku)
+		return fmt.Errorf("unpermitted jku %s", jwsHeader.Jku)
 	}
 
 	// fetch jwks (cached according to cache-control headers)
-	tp := httpcache.NewMemoryCacheTransport()
-	client := http.Client{Transport: tp}
 	resp, err := client.Get(jwsHeader.Jku)
 	if err != nil {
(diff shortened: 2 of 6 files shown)


go/examples/webhook-server/go.mod
```diff
@@ -0,0 +1,9 @@
+module webhook-server
+
+go 1.18
+
+require (
+	github.com/Truelayer/truelayer-signing/go v0.1.4 // indirect
+	github.com/gregjones/httpcache v0.0.0-20190611155906-901d90724c79 // indirect
```
tl-marco-tormento, line 7: If anyone knows about a better http cache...
[This](https://github.com/gregjones/httpcache) is not actively maintained but it does the job.

go/examples/webhook-server/main.go
```diff
@@ -0,0 +1,87 @@
...
+
+	// verify signature using the jwks
+	return tlsigning.VerifyWithJwks(jwks).Method(http.MethodPost).Path(r.RequestURI).Headers(getHeadersMap(r.Header)).Body(webhookBody).Verify(tlSignature)
+}
+
+func getHeadersMap(requestHeaders map[string][]string) map[string][]byte {
+	headers := make(map[string][]byte)
+	for key, values := range requestHeaders {
+		for _, value := range values {
+			// keep last one in case of multiple values
```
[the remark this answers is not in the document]
tl-flavio-barinas: Do we define what headers are returned by a webhook and are they standard? 

Looking at the specs, it seems that the only headers included in the request are "X-TL-Webhook-Timestamp" and "Tl-Signature" with one value each.

go/examples/webhook-server/main.go
```diff
@@ -0,0 +1,87 @@
...
+	// fetch jwks (cached according to cache-control headers)
+	resp, err := client.Get(jwsHeader.Jku)
+	if err != nil {
+		return fmt.Errorf("jku missing")
+	}
+	defer resp.Body.Close()
+	jwks, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return fmt.Errorf("jwks missing")
+	}
```
tl-flavio-barinas, line 74: does it make sense to provide this as a function within the library?

alexheretic:
Where's the readme? 

Actually related to that, you should also add links to the main go README too, like
* https://github.com/TrueLayer/truelayer-signing/blob/f7b7a258446483a0ae1145c620ac743259dc7742/csharp/README.md?plain=1#L18
* https://github.com/TrueLayer/truelayer-signing/blob/f7b7a258446483a0ae1145c620ac743259dc7742/csharp/README.md?plain=1#L41

tl-flavio-barinas (commented):
Looks good. Just a couple of questions.

go/examples/webhook-server/main.go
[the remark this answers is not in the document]
tl-flavio-barinas: Do we define what headers are returned by a webhook and are they standard?

Looking at the specs, it seems that the only headers included in the request are "X-TL-Webhook-Timestamp" and "Tl-Signature" with one value each.

go/examples/webhook-server/main.go
alexheretic: No http fetching is explicitly out of scope for these libraries, it's a rabbit hole we don't want to maintain here

tl-marco-tormento:
> Where's the readme?
That's a good question.

go/examples/webhook-server/main.go
```diff
@@ -0,0 +1,89 @@
...
+
+func verifyHook(client *http.Client, r *http.Request) error {
+	tlSignature := r.Header.Get("Tl-Signature")
+	if len(tlSignature) == 0 {
+		return fmt.Errorf("missing Tl-Signature header")
+	}
+
+	jwsHeader, err := tlsigning.ExtractJwsHeader(tlSignature)
+	if err != nil {
+		return fmt.Errorf("jku missing")
```
onatm, line 51: Why don't we return the `err` itself in here? It already has useful error message.

go/examples/webhook-server/main.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	defer r.Body.Close()
+	webhookBody, err := io.ReadAll(r.Body)
+	if err != nil {
+		return fmt.Errorf("webhook body missing")
+	}
+
+	// ensure jku is an expected TrueLayer url
+	if jwsHeader.Jku != "https://webhooks.truelayer.com/.well-known/jwks" && jwsHeader.Jku != "https://webhooks.truelayer-sandbox.com/.well-known/jwks" {
```
onatm, line 61: Shall we move this values to constants? WDYT?

go/examples/webhook-server/main.go
```diff
@@ -0,0 +1,89 @@
...
+		return fmt.Errorf("unpermitted jku %s", jwsHeader.Jku)
+	}
+
+	// fetch jwks (cached according to cache-control headers)
+	resp, err := client.Get(jwsHeader.Jku)
+	if err != nil {
+		return fmt.Errorf("jku missing")
```
onatm, line 68: I think this is more like "failed to fetch" error.

onatm (commented):
Looking good! I've just made a couple of suggestions.

go/examples/webhook-server/main.go
alexheretic: I actually tend to prefer inline for these examples, I think it makes them more "immediate" to read so conveys the intention better. But ofc it's fine either way.

go/examples/webhook-server/main.go
tl-marco-tormento: Good point!

go/examples/webhook-server/main.go
tl-marco-tormento: Yep, the curse of copy/paste has struck again...

go/examples/webhook-server/main.go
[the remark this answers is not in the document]
tl-marco-tormento: I'll take the first value.

onatm (approved):

PR #64, opened 2022-04-21
[dotnet] Fix key-dependant parameter length error for .NET Standard 2.0

tl-david-hayden:
When using the `BouncyCastle` `ECPrivateKeyParameters`, some keys `D` parameter byte representation ends up with fewer than `66` bytes, causing this error:
```
The specified key parameters are not valid. Q.X and Q.Y are required fields. Q.X, Q.Y must be the same length. If D is specified it must be the same length as Q.X and Q.Y for named curves or the same length as Order for explicit curves.
```
The fix is to prepend with padded zero value bytes to ensure a consistent byte length of `66` for `D`, `Q.X` and `Q.Y`.


tl-david-hayden:
Some context in this StackOverflow: https://stackoverflow.com/questions/48542233/translating-elliptic-curve-parameters-bc-to-ms

alexheretic (approved):

alexheretic:
Yeah I had to fix a similar issue with jwks parsing of uneven byte reprs

PR #65, opened 2022-04-29
[REC-359] provide webhook verification nodejs example

conradludgate:
Example of #47 for NodeJS


alexheretic:
Thanks! We should add a link to this example in the main nodejs readme too, like https://github.com/TrueLayer/truelayer-signing/blob/166944d77c6efa53f1494e76082b8e830fe8ba34/csharp/README.md?plain=1#L41

nodejs/examples/webhook-server/index.js
```diff
@@ -0,0 +1,72 @@
...
+    signature: sig,
+    method: req.method,
+    path: req.path,
+    headers: req.headers,
+    body: JSON.stringify(req.body),
+    jwks: JSON.stringify(jwks),
+  });
+}
+
+app.post('/hook/:id',
```
alexheretic, line 54: I think it's better to have this as the literal. In reality the webhook path is a constant, it's just the example webhook signature uses a uuid distinct path.

```suggestion
// Note: Webhook path can be whatever is configured, here a unique path 
// is used matching the README example signature.
app.post('/hook/d7a2c49d-110a-4ed2-a07d-8fdb3ea6424b',
```

conradludgate: Ah ok, sure thing

nodejs/examples/webhook-server/README.md
```diff
@@ -0,0 +1,27 @@
+# NodeJS webhook server example
+A http server than can receive and verify signed TrueLayer webhooks.
+
+## Run
+
+First, build the signing lib
+```sh
+# in the `truelayer-signing/nodejs` dir
+yarn build
```
alexheretic, line 9: Isn't it just `yarn`?

```suggestion
yarn
```

conradludgate: Maybe that works, but `build` is the specific script that does actually builds the lib.js file:
https://github.com/TrueLayer/truelayer-signing/blob/main/nodejs/package.json#L9

alexheretic: Ah I see the directory, wouldn't just running `yarn` in the examples/webhook-server dir work fine in any case?

yarn is playing up for me at the moment, but `npm i` in the webhook-server does the right thing.

alexheretic (approved):

nodejs/examples/webhook-server/README.md
conradludgate: Judging by @jdon's comment in the slack thread, it doesn't seem to work properly (it will pull the path dependency, but doesn't run the build script to build)

nodejs/examples/webhook-server/index.js
```diff
@@ -0,0 +1,79 @@
...
+  // get the (cached) JWKs for this request
+  let jwks = await get_jwks(sig);
+
+  // verify the request (will throw on failure)
+  tlSigning.verify({
+    signature: sig,
+    method: req.method,
+    path: req.path,
+    headers: req.headers,
+    body: JSON.stringify(req.body),
```
alexheretic, line 54: hmm we shouldn't really be manipulating the body in any way before verifying

PR #75, opened 2022-07-14
python: adds license + package info

tl-flavio-barinas:
releasing v0.1.1


alexheretic (approved):
symlinking the licenses would be a bit nicer, if it work for python publishing

PR #79, opened 2022-07-27
Golang: Add `Jku` method to `Signer`

tl-marco-tormento:
This PR adds `Jku` method to `Signer` to allow different jku endpoints, just like `rust` implementation.

### DIFF
--- a/go/jws/jws_header.go
+++ b/go/jws/jws_header.go
@@ -17,7 +17,7 @@ type JwsHeader struct {
 	Jku       string `json:"jku"`        // Json Web Key Url. Used in webhook signatures providing the public key jwk url.
 }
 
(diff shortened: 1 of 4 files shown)


go/jws/jws_header.go
```diff
@@ -32,6 +32,7 @@ func NewJwsHeaderV2(kid string, headers *orderedmap.OrderedMap) JwsHeader {
 		Kid:       kid,
 		TlVersion: "2",
 		TlHeaders: headerKeys,
+		Jku:       jku,
```
alexheretic, line 35: I suppose the minor downside is we'll start inflating the regular signature jwt headers with `,"jku":""`. Can we also avoid having it when empty?

alexheretic:
Ah interesting so I guess we already _were_ adding `,"jku":""` to signatures :joy:

alexheretic (approved):

Issue #80, opened 2022-07-28
Handle trailing `/` in paths more gracefully


alexheretic:
## Motivations
If a url, e.g. a webhook url, has a trailing slash it'll be signed that way `https://example.com/hook/` the path will be `/hook/`. 

When it gets to the server it's likely the path will be interpreted as `/hook`. This currently causes the verification to fail.

## Solution options
* Signer + verifier implementations both strip any trailing `/` from path. This is a simple approach and will work. However, it is technically a breaking change as old signers that have trailing `/` will not verify with new verifiers. We may be able to ensure there is no practical breakage here as I don't think this affects regular request signing, which would require no trailing slash currently, just webhook signing.
* Verifier tries a 2nd time to verify with/without a trailing `/`. This makes the verifier impls a little more complicated but is backward compatible. The extra effort to verify is probably fine particularly if it works the 2nd time.
* Ban TL webhook urls from having trailing slash. This could work, though doesn't really address the core issue in this library where I think we should.

If we can eliminate any breakage I think I'd prefer the first option.

alexheretic:
Since we already have existing trailing `/` webhook signature usage the nicer-but-breaking option 1 is going to be harder to roll out.

I think we should implement option 2:

> Verifier tries a 2nd time to verify with/without a trailing /. This makes the verifier impls a little more complicated but is backward compatible. The extra effort to verify is probably fine particularly if it works the 2nd time.

alexheretic:
Fixed in all langs

Issue #78, opened 2022-07-20
Allow for Jku endpoint setting on signer


Timunas:
## Motivations

I was trying to replicate a webhook call from truelayer to one endpoint and one of the steps on the webhook consumption as per suggested and using the `tlsigning.ExtractJwsHeader()`, the Jku endpoint is being validated as well as the signature by calling `tlsigning.VerifyWithJwks`.
For the signature bit I was using a mockserver to serve the JWKs, everything works well on signature validation however since it is not possible to set the Jku on the signer it is empty after calling `ExtractJwsHeader`, and I needed to disable the JKU endpoint validation because it won't match the mockserver expected one.
Having this possibility would allow us to use your library for both signing and validating the webhooks.

## Solution

I would like to be able to set the JKU endpoint on the signer.

## Additional context

I have tested this using golang lib.

alexheretic:
Thanks for raising this. We actually do have this [in the rust lib](https://github.com/TrueLayer/truelayer-signing/blob/main/rust/src/sign.rs#L137) and indeed this is what our webhook logic is using. It wasn't added to the other langs because it isn't necessary for regular signing.

With the test case you describe in mind it makes sense to add this to our other lang implementations, with appropriate docs that state this is for webhook signing only. It also sounds like we can solve your issue in particular by adding to golang (cc @tl-marco-tormento).

tl-marco-tormento:
Hey @Timunas, starting from version 0.1.6 of the golang lib you can now set the `jku` on the `Signer`.

PR #81, opened 2022-07-28
python:  adds path validation + jku to signer + path trailing slash

tl-flavio-barinas:
Adds and refactors to use JwsHeader type -- breaking change
Adds path validation to singer and verifier
Allows you to set the `jku` during the singing process

Bumps to `v0.2.0` from `v0.1.2`

Related issues: 
- #78 
- #69 
- #80

### DIFF
--- a/python/README.md
+++ b/python/README.md
@@ -24,7 +24,7 @@ The `verify_with_jwks` function may be used to verify webhook `Tl-Signature` hea
 
 ```python
 # `jku` field is included in webhook signatures
-let jws_header = extract_jws_header(webhook_signature)["jku"]
+jws_header = extract_jws_header(webhook_signature).jku
 
 # check `jku` is an allowed TrueLayer url & fetch jwks JSON (not provided by this lib)
 ensure_jku_allowed(jku)
--- a/python/README.md
+++ b/python/README.md
@@ -1,21 +1,13 @@
 # truelayer-signing
 
-## Description
-
 Python package to produce & verify TrueLayer API requests signatures.
 
-## How to build
+## Install
+```
+pip install truelayer-signing
+```
 
-Python: `poetry install`
-
-## Contributing
-Enable `black` and `flake-8` pre-commit hooks: `pre-commit install`
-
-## How to test
-- Type checking: `poetry run mypy ./truelayer_signing`
-
-- Test suite: `poetry run pytest -v tests/`
-
+## Generating a signature
 
 ```python
 tl_signature = sign_with_pem(KID, PRIVATE_KEY) \
--- a/python/tests/test_tl_sign.py
+++ b/python/tests/test_tl_sign.py
@@ -361,3 +361,23 @@ def test_sign_an_invalid_path():
 def test_verify_an_invalid_path():
     with pytest.raises(TlSigningException):
         verify_with_pem(PUBLIC_KEY).set_path("https://example.com/the-path")
+
+
+def test_tl_sign_set_jku():
+    body = '{"currency":"GBP","max_amount_in_minor":5000000}'
+    idempotency_key = "idemp-2076717c-9005-4811-a321-9e0787fa0382"
+    path = "/merchant_accounts/a61acaef-ee05-4077-92f3-25543a11bd8d/sweeping"
+    jku = "https://webhooks.truelayer.com/.well-known/jwks"
+
+    signature = (
+        sign_with_pem(KID, PRIVATE_KEY)
+        .set_path(path)
+        .add_header("Idempotency-Key", idempotency_key)
+        .set_body(body)
+        .set_jku(jku)
+        .sign()
+    )
+
+    jws_header = extract_jws_header(signature)
+
+    assert jws_header.jku == jku
--- a/python/tests/test_tl_sign.py
+++ b/python/tests/test_tl_sign.py
@@ -351,3 +351,13 @@ def test_verify_without_http_method():
             .add_header("X-Whatever", "aoitbeh")
             .verify(signature)
         )
+
+
+def test_sign_an_invalid_path():
+    with pytest.raises(TlSigningException):
+        sign_with_pem(KID, PRIVATE_KEY).set_path("https://example.com/the-path")
+
+
+def test_verify_an_invalid_path():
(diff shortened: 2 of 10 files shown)


python/README.md
```diff
@@ -1,21 +1,13 @@
...
-## Contributing
-Enable `black` and `flake-8` pre-commit hooks: `pre-commit install`
-
-## How to test
-- Type checking: `poetry run mypy ./truelayer_signing`
-
-- Test suite: `poetry run pytest -v tests/`
+## Install
+```
+pip install truelayer-signing
```
tl-will-tai, line 7: I see that you've removed instructions for installing using `poetry`. Given that you still have `poetry.lock` and `pyproject.toml`, do you still expect people to install using `poetry`?

python/README.md
```diff
@@ -32,7 +24,7 @@ The `verify_with_jwks` function may be used to verify webhook `Tl-Signature` hea
 
 ```python
 # `jku` field is included in webhook signatures
-let jws_header = extract_jws_header(webhook_signature)["jku"]
+let jws_header = extract_jws_header(webhook_signature).jku
```
tl-will-tai, line 27: You can remove the `let`

python/README.md
tl-flavio-barinas: Installing using poetry is not required as the lib is published and is accessible using pip https://pypi.org/project/truelayer-signing/

python/README.md
tl-flavio-barinas: good catch

python/tests/test_tl_sign.py
```diff
@@ -351,3 +351,73 @@ def test_verify_without_http_method():
...
+    (
+        verify_with_pem(PUBLIC_KEY)
+        .set_method(HttpMethod.POST)
+        .set_path("/tl-webhook")
+        .set_body(body)
+        .verify(tl_signature)
+    )
+
+
+def verify_with_unsigned_trailing_slash():
```
tl-will-tai, line 406: I'm not sure how the `verify_...` functions are used, are these functions used anywhere? Or meant to be run as unit tests?

python/truelayer_signing/__init__.py
```diff
@@ -25,20 +25,21 @@ def verify_with_pem(pkey: str) -> TlVerifier:
 
 
 def verify_with_jwks(
-    jwks: Mapping[str, List[Mapping[str, str]]], jws_header: Mapping[str, str]
+    jwks: Mapping[str, List[Mapping[str, str]]], jws_header: JwsHeader
```
tl-will-tai, line 28: Should jwks be type hinted to `Dict` instead of `Mapping`? 

https://stackoverflow.com/questions/52487663/python-type-hints-typing-mapping-vs-typing-dict

python/truelayer_signing/utils.py
```diff
@@ -21,6 +25,35 @@ class HttpMethod(str, Enum):
...
+    tl_version: str
+    tl_headers: str
+    jku: Optional[str] = None
+
+    @classmethod
+    def from_dict(cls, header: Mapping[str, str]) -> "JwsHeader":
+        if any(
+            x not in header.keys() for x in ["alg", "kid", "tl_version", "tl_headers"]
+        ):
+            raise TlSigningException("Invaild Header")
```
tl-will-tai, line 41: I think there is a typo here

python/truelayer_signing/utils.py
```diff
@@ -21,6 +25,35 @@ class HttpMethod(str, Enum):
...
+
+        if header["alg"] != "ES512":
```
tl-will-tai, line 43: As the keys `alg` and `tl_version` are always expected to be `"ES512"` and `"2"`, perhaps consider an `Enum`

tl-will-tai (commented):
Really great work, left some comments

python/tests/test_tl_sign.py
tl-flavio-barinas: updated function names with a prefix of `test`

python/truelayer_signing/utils.py
tl-flavio-barinas: fixed

python/truelayer_signing/utils.py
tl-flavio-barinas: I think creating a new type for two string literals is a bit much as they don't have any variations; I have refactored to use const instead `SIGNING_ALGORITHM and TL_VERSION`

python/truelayer_signing/__init__.py
tl-flavio-barinas: I chose `mapping` to keep the api as flexible as possible.

python/truelayer_signing/utils.py
tl-will-tai: I think I meant that because they shouldn't have variations, we can use an enum here so it's explicit as well e.g.


```
class JwsHeaderFields(Enum):
    tl_version: str = "2"
    alg: str = "ES512"

...

if header["alg"] != JwsHeaderFields.alg:

...

```

What do you think?

python/README.md
tl-will-tai: In that case, should the poetry files be removed? Or will you keep them

python/truelayer_signing/utils.py
tl-will-tai: This way, you can also use this `Enum` in the unit tests

python/README.md
tl-flavio-barinas: the poetry files are for the developers -- it's a replacement for the `setup.py` file.

python/truelayer_signing/utils.py
tl-flavio-barinas: I don't think that's an appropriate use of an `Enum`, but I do see what you're trying to do; how about this?
```python
class DefaultJwsHeaderFieldsValues:
  tl_version: str = TL_VERSION # if a new version comes out we can introduce a enum for the tl_version 
  alg: str = SIGNING_ALGORITHM # if we start to support new algorithms we can introduce a enume 
```

tl-will-tai: That makes sense!

tl-will-tai (approved):

PR #83, opened 2022-07-29
[SPAIN-367] Improve signing performances

tl-antonio-valentini:
Use the [RequestSignature](https://github.com/TrueLayer/truelayer-dotnet/blob/main/src/TrueLayer/RequestSignature.cs) class from the `truelayer-dotnet` sdk, to compare and improve the signing performances of the signing-sdk.

### TL;DR
The end user can leverage the `Signer` constructor which requires an already computed `ECDsa` (see the [comment](https://github.com/TrueLayer/truelayer-signing/pull/83#issuecomment-1199826636) below). This can be optimal in case the user wants to recycle the same key with different requests. To make things easier, we may want to turn the [ParsePem](https://github.com/TrueLayer/truelayer-signing/pull/83/files#diff-d67f01d5f39f15a2a9c2df8bb129017525d301065eeb9c4258264931e66b451bR89) method as `public` (if we don't want to make the whole `Util` class public, we can think of moving the `ParsePem` to a separate class).

### Notes
- moved the `tl-signing-test` project files inside a dedicated folder (this way we can add more test projects inside the `test` directory
- adding a basic `.editorconfig` file
- removed `netcoreapp2.2` framework because it's out of support from the test project
- added a `Benchmark` project
- made the `ParsePem` method public


tl-antonio-valentini:
Performances after 961e3a7 =>
```
// * Summary *

BenchmarkDotNet=v0.13.1, OS=ubuntu 20.04
Intel Core i9-9980HK CPU 2.40GHz, 1 CPU, 16 logical and 8 physical cores
.NET SDK=6.0.302
  [Host]     : .NET 6.0.7 (6.0.722.32202), X64 RyuJIT
  DefaultJob : .NET 6.0.7 (6.0.722.32202), X64 RyuJIT


|                Method |     Mean |     Error |    StdDev |   Median | Allocated |
|---------------------- |---------:|----------:|----------:|---------:|----------:|
|           SignWithPem | 1.022 ms | 0.0201 ms | 0.0188 ms | 1.026 ms |     11 KB |
|    SignWithPem_Static | 1.056 ms | 0.0211 ms | 0.0505 ms | 1.033 ms |     11 KB |
| SignWithPem_Static_Sb | 1.018 ms | 0.0164 ms | 0.0145 ms | 1.015 ms |     11 KB |

```

tl-antonio-valentini:
Performance after => 52c11bc
```
// * Summary *

BenchmarkDotNet=v0.13.1, OS=ubuntu 20.04
Intel Core i9-9980HK CPU 2.40GHz, 1 CPU, 16 logical and 8 physical cores
.NET SDK=6.0.302
  [Host]     : .NET 6.0.7 (6.0.722.32202), X64 RyuJIT
  DefaultJob : .NET 6.0.7 (6.0.722.32202), X64 RyuJIT


|         Method |       Mean |    Error |   StdDev |  Gen 0 | Allocated |
|--------------- |-----------:|---------:|---------:|-------:|----------:|
|    SignWithPem | 1,004.8 us | 19.33 us | 27.72 us |      - |     11 KB |
| SignWith_Ecdsa |   268.3 us |  4.97 us |  4.88 us | 0.9766 |     10 KB |
```

csharp/src/Util.cs
```diff
@@ -52,7 +52,10 @@ internal static void TryAction(Action f, string? message = null)
         internal SignatureException(string message, Exception? innerException) : base(message, innerException) { }
     }
 
-    internal static class Util
+    /// <summary>
+    /// Signing utility class.
+    /// </summary>
+    public static class Util
```
tl-mauro-franchi, line 58: If you did this just to use the `ParesePem` method in the tests, please revert it and simply allow the internal objects to be visible to the test project (see [here](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.internalsvisibletoattribute?view=net-6.0)). In this way, we will avoid exposing internal utility classes and methods to the consumers of the library.

tl-antonio-valentini: > To make things easier, we may want to turn the [ParsePem](https://github.com/TrueLayer/truelayer-signing/pull/83/files#diff-d67f01d5f39f15a2a9c2df8bb129017525d301065eeb9c4258264931e66b451bR89) method as public (if we don't want to make the whole Util class public, we can think of moving the ParsePem to a separate class).

The goal was to expose the `ParsePem` method to the end user so that it could be easier for them to leverage the `Signer.SignWith(<NewGuid>, <EcdsaKey>)`. Having it available in the test project is more of a "side-effect".

tl-antonio-valentini:
Closing this.
There's already a mechanism in place inside the truelayer-dotnet client to produce an `ECDsa` key, and we need to re-think some stuff if we really want to improve the user experience on the signing-sdk.

PR #85, opened 2022-08-11
nodejs: validate body string type

### LINKED ISSUE #84: nodejs: object body can be wrongly used without error
## Bug description

Javascript lib allow signing a object body `{foo:123}` instead of the correct string serialization `"{\"foo\":123}"` without error. However, it won't generate a usable signature so will always fail.

## Expected behavior

We should error when the incorrect type is used here.

alexheretic:
Add validation that `body` is a string. This can otherwise silently produce a useless signature.

Resolves #84

### DIFF
--- a/nodejs/src/lib.ts
+++ b/nodejs/src/lib.ts
@@ -37,6 +37,9 @@ const buildV2SigningPayload = ({
   if (!path.startsWith('/')) {
     throw new Error(`Invalid path \"${path}\" must start with '/'`);
   }
+  if (!(typeof body === 'string' || (body as any) instanceof String)) {
+    throw new Error(`Invalid body '${typeof body}' type must be a string`);
+  }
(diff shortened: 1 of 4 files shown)


nodejs/src/lib.ts
```diff
@@ -37,6 +37,9 @@ const buildV2SigningPayload = ({
   if (!path.startsWith('/')) {
     throw new Error(`Invalid path \"${path}\" must start with '/'`);
   }
+  if (!(typeof body === 'string' || (body as any) instanceof String)) {
+    throw new Error(`Invalid body '${typeof body}' type must be a string`);
+  }
```
tl-sebastien-kovacs, line 42: ❤️

tl-sebastien-kovacs (approved):
Thanks !

PR #86, opened 2022-08-15
Rust: verify handle path trailing /

alexheretic:
#80 (option 2) for rust.

### DIFF
--- a/rust/src/verify.rs
+++ b/rust/src/verify.rs
@@ -184,23 +184,20 @@ impl<'a> Verifier<'a> {
         let signing_payload =
             build_v2_signing_payload(self.method, self.path, &ordered_headers, self.body);
         let payload = format!("{header_b64}.{}", signing_payload.to_url_safe_base64());
-        let mut verify = openssl::verify_es512(&public_key, payload.as_bytes(), &signature);
-
-        if verify.is_err() {
-            // try again with/without a trailing slash (#80)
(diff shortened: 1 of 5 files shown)


rust/src/verify.rs
```diff
@@ -184,11 +183,21 @@ impl<'a> Verifier<'a> {
...
+        let payload = format!("{header_b64}.{}", signing_payload.to_url_safe_base64());
         openssl::verify_es512(&public_key, payload.as_bytes(), &signature)
-            .map_err(Error::JwsError)?;
-
-        Ok(())
+            .or_else(|e| {
+                // try again with/without a trailing slash (#80)
+                let path = match self.path {
+                    p if p.ends_with('/') => Cow::Borrowed(&p[..p.len() - 1]),
+                    p => Cow::Owned(p.to_owned() + "/"),
```
conradludgate, line 192: over-engineering hat on: should `build_v2_signing_payload` have a flag instead to write the `/` directly to the payload buffer, instead of creating an intermediate string allocation here

conradludgate: over-over-engineering hat on: why do we need the vec for the payload, when we can use https://docs.rs/openssl/0.10.41/openssl/hash/struct.Hasher.html

alexheretic: Why not :+1:

conradludgate (approved):

PR #154, opened 2023-01-11
[EWT-84] Improve release process - dotnet

Couper4:
Automate the release process by eliminating the needs for a manual release to NuGet.

The changes were tested [running the pipeline without merging the PR](https://github.com/TrueLayer/truelayer-signing/actions/runs/3912895731) in `main`: the result is a [conflicting push to NuGet](https://github.com/TrueLayer/truelayer-signing/actions/runs/3912895731/jobs/6688115243#step:5:12), which is expected since the version is the same as the one in the NuGet repository.
In a real case scenario that operation is expected to succeed.

Also adds guidelines on how to release the library in the [Wiki](https://github.com/TrueLayer/truelayer-signing/wiki/Versioning-and-Publishing#c).

### DIFF
--- a/.github/workflows/csharp.yml
+++ b/.github/workflows/csharp.yml
@@ -16,7 +16,7 @@ jobs:
   test:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
     - uses: actions/setup-dotnet@v1
       with:
         dotnet-version: '5.0.x'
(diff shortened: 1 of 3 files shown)


tl-antonio-valentini:
Should we mention the fact that people need to update the version inside [the csproj file](https://github.com/TrueLayer/truelayer-signing/blob/main/csharp/src/truelayer-signing.csproj#L3) manually? Like by putting it in the wiki?

tl-antonio-valentini (approved):

PR #167, opened 2023-03-16
Fix java7 Sonatype publishing CI

Couper4:


dili91 (approved):
LGTM - As we're touching the pipeline, feel free to also evaluate the update the gradle wrapper and the nexus publish plugin. You might have noticed some deprecation warnings while building

I'm also happy to keep these updates a separate PR but I think we'll to take care sooner than later

Couper4:
> LGTM - As we're touching the pipeline, feel free to also evaluate the update the gradle wrapper and the nexus publish plugin. You might have noticed some deprecation warnings while building
> 
> I'm also happy to keep these updates a separate PR but I think we'll to take care sooner than later

Done here: 5f725349610ffc778992b398057c2ec651491243

azanin (approved):

PR #170, opened 2023-03-24
[MAP-2] update example

tl-flavio-barinas:


adammassias1 (approved):
Thanks for adding these, would you consider doing an example for the /payments endpoint?

PR #178, opened 2023-05-11
[nodejs] Support HSM/KMS managed keys (EWT-210)

tl-david-hayden:
The library currently requires the application to have access to the private key.
Added a `sign` method overload to allow providing a function which calls an external HSM/KMS to sign with a managed key.


nodejs/src/lib.ts
```diff
@@ -176,6 +215,38 @@ export function sign(args: SignArguments) {
...
+ * @param {string} args.kid - Private key kid.
+ * @param {string} [args.method="POST"] - Request method, e.g. "POST".
+ * @param {string} args.path - Request path, e.g. "/payouts".
+ * @param {string} [args.body=""] - Request body.
+ * @param {Object} [args.headers={}] - Request headers to be signed.
+ * Warning: Only a single value per header name is supported.
+ * @returns {Promise<string>} Tl-Signature header value.
+ * @throws {SignatureError} Will throw if signing fails.
+ */
+export function signWithCallback(args: SignWithCallbackArguments): Promise<string> {
```
jdon, line 232: [nit] If we are exporting a function we should also export the type of the args. This is so our users can also use the args type in their code

tl-david-hayden: I did wonder that, but saw we don't for the existing `sign` function, so was being blindly consistent. I see we do export all the types used by the exported `verify` functions. I'll update both `sign` functions...

jdon: It seems we are generally inconsistent with what we export, we export [VerifyParameters](https://github.com/TrueLayer/truelayer-signing/blob/75fee2e110d7a780ae87473ba32684995228d714/nodejs/src/lib.ts#L296) which is used in the [verify function  ](https://github.com/TrueLayer/truelayer-signing/blob/75fee2e110d7a780ae87473ba32684995228d714/nodejs/src/lib.ts#L315) but we don't export the types for any for the signing functions.

tl-david-hayden: Actually that has made me notice we have multiple overloads of `verify` - perhaps I should rename`signWithCallback` to simply be an overload of `sign`?

jdon: I don't think we should overload, I think generally in javascript it's idiomatic for the callback to a separate argument, so I would keep it seperate and write it like this: 
```
export type signCallback = (payload: string) => Promise<string>;

export function signWithCallback(args: SignArgumentsCommon, callback: ): Promise<string> {
```

Couper4 (approved):

nodejs/src/lib.ts
tl-david-hayden: Isn't that more for callbacks that are continuations?
Perhaps me calling this a callback is confusing - it more like dependency injection, in that the function passed is used by this function to do the actual signing.

tl-david-hayden: Updated to remove the callback terminology

nodejs/README.md
```diff
@@ -39,6 +41,46 @@ const signature = tlSigning.sign({
...
+const tlSigning = require('truelayer-signing');
+
+// `Tl-Signature` value to send with the request.
+const signature = tlSigning.sign({
+  kid,
+  method: "POST",
+  path: "/payouts",
+  headers: { "Idempotency-Key": idempotencyKey },
+  body,
+  sign: async (message) => {
```
tl-david-hayden, line 59:
```suggestion
  signingFunction: async (message) => {
```

tl-peter-mills (approved):

Issue #177, opened 2023-05-10
Ruby webhook server example does not respond 202


faraquet:
## Bug description

The ruby webhook server example introduced in https://github.com/TrueLayer/truelayer-signing/pull/148 does not work properly. I am getting 401 - "Unauthorized" after doing all the steps from Readme.md

## To Reproduce
Go to ruby example folder:

```sh
cd ruby/examples/webhook-server
```

Run the server locally:

```sh
$ ruby main.rb
```

Send a test request from the doc:

```sh
curl -iX POST -H "Content-Type: application/json" \
    -H "X-Tl-Webhook-Timestamp: 2022-03-11T14:00:33Z" \
    -H "Tl-Signature: eyJhbGciOiJFUzUxMiIsImtpZCI6IjFmYzBlNTlmLWIzMzUtNDdjYS05OWE5LTczNzQ5NTc1NmE1OCIsInRsX3ZlcnNpb24iOiIyIiwidGxfaGVhZGVycyI6IngtdGwtd2ViaG9vay10aW1lc3RhbXAiLCJqa3UiOiJodHRwczovL3dlYmhvb2tzLnRydWVsYXllci5jb20vLndlbGwta25vd24vandrcyJ9..AE_QsBRhnsMkcRzd42wvY1e2HruUhkOgjuZKktGH_WmbD7rBzoaEHUuF36IxyyvCbLajd3MBExNmzjbrOQsGaspwAI5DcGVMFLKUhB7ZzUlTP9up3eNUrdwWyyfBWDQb-qmEuLnrhFDJvgCXEqlV5OLrt-O7LaRAJ4f9KHsZLQ_j2vPC" \
    -d "{\"event_type\":\"payout_settled\",\"event_schema_version\":1,\"event_id\":\"8fb9fb4e-bb2b-400b-af64-59e5dde74bad\",\"event_body\":{\"transaction_id\":\"c34c8721-66a9-49f6-a229-284efcf88a02\",\"settled_at\":\"2022-03-11T14:00:32.933000Z\"}}" \
    http://localhost:4567/hook/d7a2c49d-110a-4ed2-a07d-8fdb3ea6424b
```

Receive:
```sh
HTTP/1.1 401
Content-Type: text/plain

Unauthorized
```

## Expected behaviour

Receive 202 - Accepted

## Environment

- OS: `macOS 13.3.1`
[...]

PR #179, opened 2023-05-11
Ruby: fix webhook server example

faraquet:
### Issue
The ruby webhook server example introduced in https://github.com/TrueLayer/truelayer-signing/pull/148 does not work correctly. I am getting 401 - "Unauthorized" after doing all the steps from Readme.md

### Implementation details
In the current PR, I've restored the broken ruby webhook server example. Related to https://github.com/TrueLayer/truelayer-signing/issues/177

The reason was incorrect `TrueLayerSigning` settings, and incorrect parsing of request headers.

- Introduced `ensure_certificate_id_present!` method which checks the certificate is present or assigns a random one. This fixed `TRUELAYER_SIGNING_CERTIFICATE_ID is missing` error from [truelayer-signing/ruby/lib/truelayer-signing/utils.rb](https://github.com/TrueLayer/truelayer-signing/blob/b64bb284f58b2d6a86622d18943c585744ff8ef3/ruby/lib/truelayer-signing/utils.rb#L6).
- Server moved to `webrick` to get convinient access to headers, path, and body.
- In the previous implementation, the `x-tl-webhook-timestamp`  header was parsed to incorrect value  `33Z`, leading to the failing signature verification.

### DIFF
--- a/ruby/examples/webhook-server/Gemfile
+++ b/ruby/examples/webhook-server/Gemfile
@@ -1,3 +1,4 @@
 source "https://rubygems.org"
 
 gem "truelayer-signing"
+gem "webrick"
--- a/ruby/examples/webhook-server/main.rb
+++ b/ruby/examples/webhook-server/main.rb
@@ -1,5 +1,5 @@
-require 'truelayer-signing'
 require 'securerandom'
+require 'truelayer-signing'
 require 'webrick'
 
 class TrueLayerSigningExamples
--- a/ruby/examples/webhook-server/main.rb
+++ b/ruby/examples/webhook-server/main.rb
@@ -17,8 +17,7 @@ class TrueLayerSigningExamples
 
   class << self
     def run_webhook_server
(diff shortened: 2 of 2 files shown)


faraquet:
@kplattret could you please have a look?

kplattret (commented):
Thanks for spotting this issue and raising a PR. I've left a few comments but other than that it looks good.

kplattret:
Also, you need to add the following to the `Gemfile` as it doesn't seem to come with Ruby by default:

```ruby
gem "webrick"
```

faraquet:
Thanks for the detailed review @kplattret ! I fixed all your comments.
I was pretty sure the `webrick` it's part of the ruby stdlib, but as I can see now since version 3 they (seem to) have removed it from there https://bugs.ruby-lang.org/issues/17303

kplattret (approved):
Thanks @faraquet. I'll get this out in a new release later today or tomorrow.

PR #183, opened 2023-05-18
Ruby: fix conflict with JWT library

kplattret:
The custom code patch we have for the `jwt` library, which prevents the payload object to be systematically parsed as JSON, breaks the original `JWT.encode` and `JWT.decode` methods. That's a problem for applications already using `JWT` for other purposes, since adding `truelayer-signing` suddenly causes unexpected errors.

This introduces the `JWT.truelayer_encode` and `JWT.truelayer_decode` methods, which allow us to modify the behaviour without compromising the underlying functionality. At some point we should propose a change to the `JWT` library and remove this custom code patch.

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -26,6 +26,6 @@ jobs:
         with:
           ruby-version: ${{ matrix.ruby-version }}
       - name: Build and install gem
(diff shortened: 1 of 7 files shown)


faraquet:
Thanks @kplattret 👍

conradludgate (approved):

PR #186, opened 2023-06-07
Ruby: add verify_with_jwks method

kplattret:
Adds support for signature verification using JWKs. Introduces
`TrueLayerSigning.extract_jws_header(signature)` for convenience and
`TrueLayerSigning.verify_with_jwks(jwks)` to verify signatures more
easily, without having to deal with generating the PEM manually.

When any EC key coords value is shorter than 66 bytes, this ensures that
the correct number of zero values are prepended to the coords, making
them valid and allowing the public key to be imported correctly.

Jira ticket: https://truelayer.atlassian.net/browse/RSK-592


kplattret:
The new features are pending some additional tests.

tl-flavio-barinas (approved):
LGTM

conradludgate (approved):

ruby/README.md
```diff
@@ -59,10 +59,20 @@ See full example of [request signing].
...
-The `verify_with_pem` method may be used to verify webhook `Tl-Signature` header signatures.
+The `verify_with_jwks` method may be used to verify webhook `Tl-Signature` header signatures.
 
 ```ruby
-TrueLayerSigning.verify_with_pem(pem)
+# The `jku` field is included in webhook signatures
+jku = TrueLayerSigning.extract_jws_header(webhook_signature).jku
+
+# You should check that the `jku` is a valid TrueLayer URL (not provided by this library)
+ensure_jku_allowed(jku)
```
lboynton, line 69: Is there a reason for not doing this as part of the library?

lboynton: Oh I see, the SDKs do that. Ignore me.

lboynton (approved):

ruby/README.md
kplattret: I'm not sure why we decided not to build this into the libraries, however we do provide an example of how it can be achieved.

PR #189, opened 2023-06-22
Ruby: add Missing header discovery

GeoffTidey:
Added some helper code to determine which header is missing 

eg./ I discovered receiving `HTTP_TL_WEBHOOK_TIMESTAMP` when `tl-webhook-timestamp` is wanted.

Also added Gemfile to include minitest for the tests

### DIFF
--- a/ruby/Gemfile
+++ b/ruby/Gemfile
@@ -1,0 +1,5 @@
+source 'https://rubygems.org'
+
+group :development, :test do
+  gem 'minitest'
+end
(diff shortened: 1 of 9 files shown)


ruby/Gemfile
```diff
@@ -0,0 +1,5 @@
+source 'https://rubygems.org'
+
+group :development, :test do
+  gem 'minitest'
+end
```
kplattret, line 5: This isn't needed since `minitest` is a core gem that comes with Ruby, I believe. I've never needed to install it for testing our gem, neither locally nor in the GitHub action.

kplattret (approved):

PR #195, opened 2023-07-11
Ruby: disable expiration verification

kplattret:
Verification of expired-payment webhook signatures is failing. This is
because the underlying `jwt` library expects the payload to be a JSON
object, but ours isn't, as per our request signing spec. This breaks
some code paths and the expiration verification was missed by our
manual/automated tests.

This disable expiration verification, as well as any other verification
that expects the payload to be a JSON object.

Jira ticket: https://truelayer.atlassian.net/browse/RSK-647


adamteddychang:
It works @kplattret , nice work 👍

tl-marco-tormento (approved):

PR #215, opened 2023-11-10
DOCS-973 Update request-signing-v2.md

t-j-harrington:
Updating key generation bash commands to use openssl rather than Docker following feedback/confusion related to old commands. Also updated the wording/link for Console.

### DIFF
--- a/request-signing-v2.md
+++ b/request-signing-v2.md
@@ -17,16 +17,16 @@ You can generate a key pair using [`openssl`](https://www.openssl.org/).
 To generate the private key, run:
 
 ```bash
-docker run --rm -v ${PWD}:/out -w /out -it alpine/openssl ecparam -genkey -name secp521r1 -noout -out ec512-private-key.pem
+openssl ecparam -genkey -name secp521r1 -noout -out ec512-private-key.pem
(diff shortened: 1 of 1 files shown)


request-signing-v2.md
```diff
@@ -17,16 +17,16 @@ You can generate a key pair using [`openssl`](https://www.openssl.org/).
 To generate the private key, run:
 
 ```bash
-docker run --rm -v ${PWD}:/out -w /out -it alpine/openssl ecparam -genkey -name secp521r1 -noout -out ec512-private-key.pem
+openssl ecparam -genkey -name secp521r1 -noout -out ec512-private-key.pem
```
kplattret, line 20: This could potentially cause issues as different systems might use a different version of OpenSSL, whereas using Docker guarantees some consistency. I'm curious to know what feedback we got around this exactly.

t-j-harrington: Made a change to follow approach proposed above.

kplattret: I think it's great to list both options indeed. Thank you!

kplattret (approved):

PR #219, opened 2023-11-29
Ruby: add code linter

kplattret:
This introduces the use of Rubocop to ensure that the code-style is
consistent across the gem's source code. It also adds it as a CI step.


kplattret:
I have also tested both the `sign-request` and `webhook-server` examples, and I can confirm they still work fine after these changes. 👍

tl-flavio-barinas (approved):
Not a ruby expert but LGTM

PR #230, opened 2023-12-05
Python: publish on version bump only

kplattret:
Since #228, the Python workflow will run for any changes in the Python
code, the shared test resources _and_ the Python workflow itself. This
means that when the library version does not change, the `publish` step
simply fails, which does not look good.

This updates the Python workflow so that it will only attempt to publish
the library automatically when there is a version bump on `main`.

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -53,6 +53,8 @@ jobs:
 
   check_version:
     runs-on: ubuntu-latest
+    # Runs only on the main branch
+    if: ${{ github.ref == 'refs/heads/main' }}
     needs: lint_and_test
     outputs:
       version: ${{ steps.get_version.outputs.VERSION }}
@@ -71,8 +73,8 @@ jobs:
 
   publish:
     runs-on: ubuntu-latest
-    # if: ${{ github.ref == 'refs/heads/main' && needs.check_version.outputs.is_published == false }}
-    if: ${{ needs.check_version.outputs.is_published == false }}
+    # Runs only on the main branch and if the current version isn't already published
+    if: ${{ github.ref == 'refs/heads/main' && needs.check_version.outputs.is_published == false }}
     needs: [lint_and_test, check_version]
     env:
       python-version: "3.10"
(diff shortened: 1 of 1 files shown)


tl-flavio-barinas (commented):
LGTM, although I would like to avoid the manual step.
How about a script that checks if the version exists and leaves a warning annotation?

```bash
if ! VERSION="$(grep -m1 'version = "' pyproject.toml | cut -d'"' -f2)"; then exit 1; fi
STATUS_CODE=$(curl --write-out %{http_code} --silent --output /dev/null "https://pypi.org/pypi/truelayer-signing/${VERSION}/json")

if [ $STATUS_CODE -eq 200 ] ; then
    echo "::warning truelayer-signing:$VERSION::Version already exists skipping publication."
else
    poetry build
    poetry config pypi-token.pypi ${{ secrets.PYPI_API_TOKEN }}
    poetry publish
fi
```

kplattret:
@tl-flavio-barinas I don't mind at all if that's your preference. I think this should work well for whoever maintains the specific library really.

The one thing worth pointing out is that skipping the `publish` job altogether is clearer from a workflow perspective. As shown on the screenshot below, it's more reassuring to see that the library was not published, rather than see the step as successfully executed, then having to check or remember that it didn't actually work as there was no version bump.

image

tl-flavio-barinas:
> @tl-flavio-barinas maybe we can find a way of doing your check in the job's conditional statement though, as opposed to doing it in an execution step?

It would be nice if we could filter at a job level. Not sure it's possible, though. We could rename the job to `try_publish` + the annotation + update the script to fail if the code has changed and the version hasn't.

kplattret:
@tl-flavio-barinas I've updated the workflow as such:

1. Added a `check_version` job that will only run when pushing to the `main` branch and if the `lint_and_test` job succeeds. You can see it work [here](https://github.com/TrueLayer/truelayer-signing/actions/runs/7144307579/job/19457651054).
2. Updated the `publish` job so that it only runs:
    - when pushing to the `main` branch;
    - if `lint_and_test` and `check_version` succeed;
    - if current version's `is_published` is `false`.

This removes the manual step of tagging and keeps the workflow elegant, with the `publish` job only running when necessary, instead of going red.

tl-flavio-barinas (approved):
LGTM

PR #237, opened 2023-12-11
Python: release new patch version

kplattret:
There have been a couple of dependency updates since the last release,
so this bumps up the patch version. The revised workflow should pick up
the new package version and publish it automatically.


tl-flavio-barinas (approved):

tl-flavio-barinas (approved):
LGTM, small nit, the change log needs updating

PR #253, opened 2024-02-16
[PAYINS-464] [csharp] Add capability to sign with function

tl-wajid-malik:
Add `Signer.SignWithFunction(string kid, Func<string, Task<string>> signAsync)`  to allow consumers to pass a function which takes the signing payload and returns the signature. This allows signing to be delegated to an external key management service in the case that the library consumer does not have direct access to the private key.

The changes have been implemented in a non-breaking way for existing consumers.

Related PR for JS implementation: https://github.com/TrueLayer/truelayer-signing/pull/178

### DIFF
--- a/csharp/src/Signer.cs
+++ b/csharp/src/Signer.cs
@@ -12,7 +12,7 @@ namespace TrueLayer.Signing
     /// <summary>
     /// Builder to generate a Tl-Signature header value using a private key.
     /// </summary>
-    public sealed class Signer
+    public sealed class Signer : Signer<Signer>
     {
         /// <summary>
         /// Start building a request Tl-Signature header value using private key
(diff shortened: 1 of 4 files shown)


csharp/src/Signer.cs
```diff
@@ -132,27 +153,79 @@ public Signer Body(byte[] body)
...
+
+            var headerList = _headers.Select(e => (e.Key, e.Value)).ToList();
+            var signingPayload = Util.BuildV2SigningPayload(_method, _path, headerList, _body);
+            var signingPayloadB64 = Base64Url.Encode(signingPayload);
+
+            var signingMessage = $"{serializedJwsHeadersB64}.{signingPayloadB64}";
+
+            var signature = await _signAsync(signingMessage);
+
+            return $"{serializedJwsHeadersB64}..{signature}";
```
tl-tai-tang, line 228: Is this additional `dot` intentional? For clients to insert the _body in the middle?

tl-tai-tang (approved):

csharp/src/Signer.cs
tl-wajid-malik: Yes here it's intentional. The standard JWS has the format `<headers>.<payload>.<signature>`, and the detached JWS which we use for our signatures have the payload removed, leaving the format `<headers>..<signature>`.

PR #266, opened 2024-03-11
rust: allow reponse GET body sign/verify

tl-flavio-barinas:
The Body will no longer be ignored for requests method GET or DELETE.


Lindronics (commented):
Looks good. Can we populate body by default, so we don't need to set it explicitly for get?

Lindronics (approved):

PR #280, opened 2024-04-24
[rust] Support providing tl_version + tl_headers via HTTP headers

Haydabase:


rust/src/verify/mod.rs
```diff
@@ -267,12 +268,12 @@ impl<'a> Verifier<'a> {
         let parsed_tl_signature = parse_tl_signature(tl_signature)?;
 
         match &parsed_tl_signature.header.tl_version {
-            Some(tl_version) if tl_version == "1" => VerifierV1 {
+            None | Some(TlVersion::V1) => VerifierV1 {
```
Haydabase, line 271: Why do we want `None` to be parsed as V1?

rust/src/lib.rs
```diff
@@ -30,10 +31,10 @@ pub use verify::{CustomVerifier, Verifier, VerifierBuilder};
 ///     .sign()?;
 /// # Ok(()) }
 /// ```
-pub fn sign_with_pem<'a>(
-    kid: &'a str,
-    private_key_pem: &'a [u8],
-) -> SignerBuilder<'a, &'a str, &'a [u8], Unset, Unset, Unset> {
+pub fn sign_with_pem(
+    kid: Uuid,
```
Haydabase, line 35: This is a breaking change to the library right?

rust/src/verify/mod.rs
tl-flavio-barinas: If no version was provided, we assumed that it would be v1. 
https://github.com/TrueLayer/truelayer-signing/blob/b752b13309cfa27bb8bf367b46a3189bd141f0f0/rust/src/verify/mod.rs#L275

rust/src/lib.rs
Haydabase: Do we specify the KID will always be a UUID in our signing requirements?
Is there a good reason to make this change to the library, especially given other language libraries treat it as an opaque string?

rust/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "truelayer-signing"
-version = "0.2.1"
+version = "0.2.2"
```
tl-flavio-barinas, line 3: This should be a breaking change due to the types being changed. 
```suggestion
version = "0.3.0"
```

tl-flavio-barinas (approved):

azanin (approved):

PR #285, opened 2024-05-13
[PAYINS-600] [csharp] Allow specifying jku when creating signature

tl-wajid-malik:
Allow specifying a `jku` when creating a signature, which will be included in the JWS headers.
This is intended to be used for response signing and will point consumers to TrueLayer's JWKS for validating these signatures.


suryapandian (commented):
@tl-wajid-malik you could also update [CODEOWNER](https://github.com/TrueLayer/truelayer-signing/blob/main/.github/CODEOWNERS) file accordingly. seems like only rustoss is requested for this PR

PR #294, opened 2024-06-16
[EWT-581] UTF-8 old fix release + Java7 library update

dili91:
Apparently [this old fix](https://github.com/TrueLayer/truelayer-signing/pull/192) was never released. 
Plus, we need the same to happen on the java7 version of the lib.

This PR also update the gradle wrapper and the CI as they were outdated


.github/workflows/java.yml
```diff
@@ -23,7 +23,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ ubuntu-latest ]
-        java: [ 8, 9, 10, 11, 12, 13, 14, 15, 16, 17 ]
+        java-version: [ 8, 11, 17, 20, 21 ]
```
dili91, line 26: I'm removed some old setups and added new ones. This mimics the setup we have for truelayer-java

java/gradle/wrapper/gradle-wrapper.jar
dili91, line 1: This is and the other gradle-specific updates are simply consequence of running `gradle wrapper` command to update the wrapper

java7/lib/src/main/java/com/truelayer/signing/Utils.java
```diff
@@ -45,7 +45,7 @@ public static Base64URL buildPayload(Map<HeaderName, String> headers, String met
                 path +
                 "\n" +
                 headerStringBuilder +
-                new String(body);
+                new String(body, StandardCharsets.UTF_8);
```
dili91, line 48: I've ported the old fix also on the java7 lib

azanin (approved):

PR #372, opened 2025-11-23
perf(PAYINS-1388): Add native AOT support

tl-Roberto-Mancinelli:
## Summary

This PR adds Native AOT (Ahead-of-Time) compilation support to the TrueLayer.Signing library, enabling better performance and reduced startup times for .NET applications.

### Key Changes

- **Removed Jose.JWT dependency**: Replaced with custom implementation to eliminate reflection-based code
- **Added JSON Source Generation**: Implemented `SigningJsonContext` for AOT-compatible JSON serialization
- **Custom Base64Url encoding**: Implemented custom Base64Url encoder/decoder without reflection
- **Custom JWS verification**: Replaced Jose.JWT's verification logic with direct cryptographic operations
- **Target framework dependencies**: Split dependencies by target framework (net10.0, net9.0, net8.0, netstandard2.0, netstandard2.1)
- **AOT annotations**: Added `IsAotCompatible`, `IsTrimmable`, `EnableTrimAnalyzer`, and `EnableSingleFileAnalyzer` properties
- **Fixed XML documentation warnings**: Added missing XML comments for public API members

### Benchmark (net10/small payment)
current:
| Method           | Mean     | Error    | StdDev   | Rank | Gen0   | Allocated |
|----------------- |---------:|---------:|---------:|-----:|-------:|----------:|
| 'Verify Request' | 667.5 us | 12.92 us | 13.82 us |    1 | 0.9766 |   9.13 KB |        
| 'Sign Request'   | 677.5 us | 12.80 us | 14.22 us |    1 |        |   7.12 KB | 

v0.2.5 - jose-jwt:
| Method           | Mean     | Error   | StdDev  | Rank | Gen0   | Allocated |
|----------------- |---------:|--------:|--------:|-----:|-------:|----------:|
| 'Verify Request' | 665.5 us | 4.21 us | 3.52 us |    1 | 0.9766 |  14.73 KB |
| 'Sign Request' | 677.6 us | 10.55 us | 11.29 us |    1 | 0.9766 |  13.47 KB | 

v0.1.16
| Method           | Mean     | Error    | StdDev   | Rank | Gen0   | Allocated |
|----------------- |---------:|---------:|---------:|-----:|-------:|----------:|
| 'Verify Request' | 679.9 us | 13.26 us | 16.28 us |    1 | 1.9531 |  21.56 KB | 
| 'Sign Request'   | 682.2 us | 13.60 us | 13.36 us |    1 | 0.9766 |  13.67 KB |


csharp/src/Verifier.cs
```diff
@@ -277,20 +320,74 @@ public void Verify(string tlSignature)
...
+                // Compute SHA-512 hash of the signing input (ES512 uses SHA-512)
+#if NET5_0_OR_GREATER
+                var hash = SHA512.HashData(signingInput);
+#else
+                byte[] hash;
+                using (var sha512 = SHA512.Create())
+                {
+                    hash = sha512.ComputeHash(signingInput);
+                }
+#endif
```
tl-wajid-malik: This code seems shared between the Signer and Verifier, could we extract to a Util?

tl-wajid-malik: Also we differ between ASCII and UTF8 encoding in the if block, why is that?

tl-Roberto-Mancinelli: the output in this case is the same as base64url is producing ASCII only but better to align both paths

PR #416, opened 2026-06-17
Add minimal GITHUB_TOKEN permissions to workflows

kplattret:
## Summary

Adds explicit, least-privilege `permissions` blocks to the workflows that lacked one, fixing the **`actions/missing-workflow-permissions`** code-scanning alerts. Restricting `GITHUB_TOKEN` to the minimum needed limits the blast radius if a step or dependency is compromised.

## Changes

**Top-level `permissions: contents: read`** — read-only is sufficient (checkout only; publishing uses dedicated secret tokens, not GitHub write scopes):

| Workflow | Publishing mechanism |
|---|---|
| `csharp.yml` | NuGet (`NUGET_API_KEY`) |
| `go.yml` | — |
| `node.yml` | npm (`NPM_TOKEN`) |
| `php.yml` | downstream sync via SSH `deploy_key` |
| `python.yml` | PyPI (`PYPI_API_TOKEN`) |
| `ruby.yml` | RubyGems (`GEM_HOST_API_KEY`) |

**Top-level `contents: read` + a job-level write grant** — these have release jobs that genuinely need write access, so a blanket `read` would break them:

- `java.yml` / `java7.yml` — the `create_github_release` job gets `contents: write` (git tag via `mathieudutour/github-tag-action` + release via `softprops/action-gh-release`).
- `rust.yml` — the `publish` job gets `contents: write` + `pull-requests: write` (`release-plz` opens release PRs and creates tags/releases).

A job-level `permissions` block fully overrides the top-level one, so each release job lists exactly the scopes it needs while every other job inherits the safe `contents: read`.

**Not changed:** `codeql.yml` already defines `permissions` on both of its jobs, so no alert fires there.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -91,6 +91,13 @@ jobs:
     name: Analyze (${{ matrix.language }})
     needs: detect_changes
     # Skip entirely when no relevant language changed (empty matrix).
+    #
+    # Because PRs only analyse changed languages, the code scanning check may
+    # report "configurations not found" / a neutral result: it can't diff the
+    # languages present on main but not re-run here. This is expected and
+    # harmless — unchanged languages can't introduce new alerts, and main still
+    # analyses everything on push. Avoid requiring a "success" (vs neutral)
+    # conclusion for this check in branch protection, or such PRs would block.
     if: needs.detect_changes.outputs.matrix != '[]'
     runs-on: ubuntu-latest
     permissions:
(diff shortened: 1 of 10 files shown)


kplattret:
The build/test jobs for every language (`csharp`, `go`, `java`, `java7`, `nodejs`, `php`, `python`, `ruby`, `rust`) ran with the new `contents: read permission` and passed. This confirms that the tightened token scope didn't break anything.

lboynton (approved):

PR #421, opened 2026-06-19
java: update deps, test on each LTS runtime, and release 0.3.0

kplattret:
Updates the `/java` library's dependencies and build to the latest versions while keeping the **published library Java 8 compatible** (bytecode major version 52), so existing consumers are unaffected. Releases as **0.3.0**.

## Library (stays Java 8)
- `com.nimbusds:nimbus-jose-jwt` 9.37.3 → **10.9.1**. Nimbus 10.x is a multi-release JAR with a Java 7 base, verified to run on Java 8.
- Migrated `org.bouncycastle:bcpkix-jdk15on:1.70` → **`bcpkix-jdk18on:1.84`** — the `jdk15on` artifact line is end-of-life/deprecated; `jdk18on` still targets Java 8+.
- Gradle wrapper 8.3 → **8.14.5** (+ `distributionSha256Sum`). Stays on 8.x so the build tooling still runs on the older JDKs.
- Updated two test assertions: Nimbus 10.x changed its malformed-JSON error message wording; library behaviour is unchanged.

## webhook-server example (only consumes the Java 8 library)
- `io.javalin:javalin` 4.4.0 → **7.2.2**, example toolchain → **Java 21** (Javalin 7 runs on Jetty 12 / Jakarta EE 10, which requires Java 17+).
- Rewrote route/exception registration for Javalin's config-based API.

## CI: genuinely test on each LTS runtime
Previously the matrix only changed which JDK ran Gradle — because the lib toolchain is pinned to Java 8, the tests always executed on Java 8. Now the test task's launcher follows `-PtestJavaVersion`, so the suite is exercised on each LTS runtime.
- Matrix updated to the current LTS set: **8, 11, 17, 21, 25** (dropped EOL/non-LTS 20, added LTS 25).
- `setup-java` installs Java 8 (compile toolchain), the matrix version (test runtime) and Java 21 last so it runs Gradle (Gradle 8.x can't run on Java 25+, but toolchains let us still test on it).
- Replaced the archived `gradle/wrapper-validation-action` with the maintained `gradle/actions/wrapper-validation@v6.2.0`.
- The `test` task logs the three JVMs it uses at startup — compile toolchain, Gradle daemon (`JAVA_HOME`), and test runtime — so each matrix job's log makes the separation auditable rather than implicit.

## Release
Bumps `version` to **0.3.0** and updates the CHANGELOG (also backfills the missing 0.2.6 entry). Minor bump because the runtime dependency tree changes materially (Nimbus major bump; BouncyCastle artifact migration), even though the public API is unchanged.

> ⚠️ Merging this to `main` will trigger the automatic Maven Central publish + `java-v0.3.0` tag/release, since `gradle.properties` version changed.

## Verification
- Full clean build passes across both toolchains (lib + both examples, Javadoc/sources JARs).
- All five CI matrix legs (8, 11, 17, 21, 25) are green, and the per-job logs confirm the JVM split. Examples: the **Java 25** leg compiles on `1.8.0_492`, runs Gradle on `21.0.11` (`JAVA_HOME` = Temurin 21), and runs tests on `25.0.3+9-LTS`; the **Java 8** leg compiles and tests on `1.8.0_492`. Compile is always Java 8, Gradle is always 21, only the test runtime tracks the matrix.
- webhook-server example boots on Jetty 12 / Java 21 and serves requests against the Java 8 library (POST without `Tl-Signature` → 400).

### DIFF
--- a/.github/workflows/java.yml
+++ b/.github/workflows/java.yml
@@ -26,22 +26,28 @@ jobs:
       fail-fast: false
       matrix:
         os: [ ubuntu-latest ]
-        java-version: [ 8, 11, 17, 20, 21 ]
+        java-version: [ 8, 11, 17, 21, 25 ]
         java-distribution: [ temurin ]
     runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout current branch
         uses: actions/checkout@v7
-      - name: Setup JDK
+      # Install three JDKs: 8 for the compile toolchain (the library is always built
+      # to Java 8 bytecode), the matrix version as the test runtime, and 21 last so it
+      # becomes JAVA_HOME and runs Gradle (Gradle 8.x cannot run on Java 25+).
+      - name: Setup JDKs
         uses: actions/setup-java@v5
         with:
-          java-version: ${{ matrix.java-version }}
           distribution: ${{ matrix.java-distribution }}
+          java-version: |
+            8
+            ${{ matrix.java-version }}
+            21
           cache: 'gradle'
       - name: Validate gradle wrapper
         uses: gradle/wrapper-validation-action@f9c9c575b8b21b6485636a91ffecd10e558c62f6 # v3.5.0
-      - name: Test
-        run: cd java && ./gradlew test
+      - name: Test on Java ${{ matrix.java-version }}
+        run: cd java && ./gradlew test -PtestJavaVersion=${{ matrix.java-version }}
 
   set-release-version:
     name: Prepare version to tag and release
@@ -75,11 +81,14 @@ jobs:
     needs: [build-and-test, set-release-version]
     steps:
       - uses: actions/checkout@v7
-      - name: Setup JDK
+      # Java 8 for the compile toolchain; 21 last so it runs Gradle.
+      - name: Setup JDKs
         uses: actions/setup-java@v5
         with:
-          java-version: '21'
           distribution: 'temurin'
+          java-version: |
+            8
+            21
           cache: 'gradle'
       - name: Validate gradle wrapper
         uses: gradle/wrapper-validation-action@f9c9c575b8b21b6485636a91ffecd10e558c62f6 # v3.5.0
--- a/.github/workflows/java.yml
+++ b/.github/workflows/java.yml
@@ -45,7 +45,7 @@ jobs:
             21
           cache: 'gradle'
       - name: Validate gradle wrapper
-        uses: gradle/wrapper-validation-action@f9c9c575b8b21b6485636a91ffecd10e558c62f6 # v3.5.0
+        uses: gradle/actions/wrapper-validation@3f131e8634966bd73d06cc69884922b02e6faf92 # v6.2.0
       - name: Test on Java ${{ matrix.java-version }}
         run: cd java && ./gradlew test -PtestJavaVersion=${{ matrix.java-version }}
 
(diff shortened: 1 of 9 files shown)


.github/workflows/java.yml
```diff
@@ -26,22 +26,28 @@ jobs:
...
         with:
-          java-version: ${{ matrix.java-version }}
           distribution: ${{ matrix.java-distribution }}
+          java-version: |
+            8
+            ${{ matrix.java-version }}
+            21
           cache: 'gradle'
       - name: Validate gradle wrapper
         uses: gradle/wrapper-validation-action@f9c9c575b8b21b6485636a91ffecd10e558c62f6 # v3.5.0
```
dili91: This project is archived. more info [here](https://github.com/gradle/wrapper-validation-action)

.github/workflows/java.yml
```diff
@@ -26,22 +26,28 @@ jobs:
...
         os: [ ubuntu-latest ]
-        java-version: [ 8, 11, 17, 20, 21 ]
+        java-version: [ 8, 11, 17, 21, 25 ]
         java-distribution: [ temurin ]
     runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout current branch
         uses: actions/checkout@v7
-      - name: Setup JDK
+      # Install three JDKs: 8 for the compile toolchain (the library is always built
```
dili91, line 35: Does it makes sense to have 2 separate workflows: one for the ancient java 8, and another one for more recent versions? We could DRY (if we want to) with some [reusable parts maybe](https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows)

kplattret: The library is on Java 8 for maximum compatibility, so it needs to install that along with each Java version we test against in this workflow. I don't think we can remove Java 8 for other versions, since it's needed to build the library.

Unless I misunderstood your question?

.github/workflows/java.yml
kplattret: https://github.com/TrueLayer/truelayer-signing/pull/421/commits/5b1161063de0c8a67cfda1f75610cbf0e55fa963

.github/workflows/java.yml
dili91: By looking at the java-version change below and this comment  

> (Gradle 8.x cannot run on Java 25+)

I thought CI broke if java 25 was used. I guess I'm simply not getting the need of amending the java-versions matrix below (and also I'm a bit puzzled by gradle's compatibility matrix). To reduce the ambiguity, do you have a link to a failing workflow?

Unless I'm misunderstanding [CI succeeded](https://github.com/TrueLayer/truelayer-signing/actions/runs/27832220174/job/82371554104) on java 25?

kplattret: So this is what I believe this change is doing:

- Java 8 is needed for each version, to build the library, which uses it (as the oldest Java version still being maintained) for maximum compatibility.
- Java 21 is needed because Gradle 8.x needs it.
- Then we test it all with each currently maintained version of Java, including 25.

Does that help clarify things? Previously we were not actually testing with each version, just Java 8, so this is a bit of a departure.

java/examples/webhook-server/build.gradle
```diff
@@ -8,12 +8,12 @@ repositories {
 
 java {
     toolchain {
-        languageVersion = JavaLanguageVersion.of(8)
+        languageVersion = JavaLanguageVersion.of(21)
```
kplattret, line 11: The code example can use Java 21. It's still compatible with the library and it's best to use the latest LTS really.

.github/workflows/java.yml
dili91: Discussion is continuining on Slack but I don-t think there's any point blocking this further. we can continue discussing and defer reworks

dili91 (approved):

